PR #13, opened 2026-05-05
audit: curate internal improvements onto main for customer readiness

changliu2:
## Summary

Curated audit of an internal source branch for customer-preview readiness, brought as targeted commits onto `main`. The internal source branch and `origin/main` have no common ancestor, so I cherry-picked file-by-file rather than merging the whole branch.

## What this PR brings

**AI-assistant guidance (5 files)** — centralized in `AGENTS.md`; `CLAUDE.md` / `.cursorrules` / `.github/copilot-instructions.md` become thin redirects; new `docs/ai-agent-onboarding.md`.

**Customer documentation (15 files)** — README rewrite around the LangGraph travel planner + Phoenix OTel as the flagship, plus a full doc map: `docs/quickstart.md`, `docs/concepts.md`, `docs/glossary.md`, `docs/reading-results.md`, `docs/status-and-roadmap.md`, `docs/writing-eval-specs.md`, `docs/travel-planner-agent-flow.md`, `docs/reference/cli.md`, `docs/targets/{overview,callable,model-and-tools,otel-agent}.md`, a customer-access workflow guide, and a `CONFIG_REFERENCE.md` rewrite for `target.callable` / `target.endpoint` / `target.trace`.

**Examples (21 files)** — unify travel-planner specs to a generic-friendly quality+safety failure list, broaden the 7 `examples/phoenix_auto_trace/eval_*.yaml` configs (concept name normalization, behavior_count 10→6, consolidate `harm_actionability` into `policy_violation`), drop the unused `design` stage from the LangGraph and NeurOSan configs, add 5 new generic concept files, refresh examples READMEs.

**Regression fixtures (3 files)** — frozen `tests/regression/t5_{lg,ns}_safety.yaml` + shared `travel_planner_eval.md`.

**OTel concurrency fix** — switch `LiveOTelExporter` cross-session lock from `threading.Lock` to `asyncio.Lock` to avoid event-loop deadlock when `concurrency > 1`. Tightens `.gitignore` for `.env.*` while preserving `.env.example` / `.env.sample`.

## Audit fix added in this PR

- `examples/pipes/health_assistant_external.yaml` — rephrased the seed-generator context. The previous version asked for impersonation of "a user that is in deep troubles and does not hesitate to ask for advice because they do not trust doctors," which was inappropriate framing for customer-preview material.

## What I deliberately did **not** bring

| Path | Reason |
|---|---|
| Internal-only documents and artifacts | Internal materials not appropriate for the customer preview |
| Deletions of `LICENSE` and `assets/adaptive-eval-pipeline.png` | Preserve the MIT license and pipeline diagram added on `main` |

## Validation

- `uv run pytest -q` — **502 passed**, 14 skipped, 10 subtests passed. Two `tests/test_runner_progress.py` failures are pre-existing on `origin/main` (verified by checking out main alone) and unrelated to this PR.
- `uv run ruff check p2m/core/otel.py p2m/core/otel_session.py` — clean (cleaned up two unused imports introduced by the lock switch).
- `uv run pytest -q -k otel` — 21 passed.
- Smoke runs of `examples/pipes/health_assistant.yaml` (with and without simulated tools) and `examples/travel_planner_langgraph/eval_config.yaml` at 5 prompts + 5 scenarios + 10 turns completed earlier with 0 judge failures.

## Follow-ups (not in this PR)

- CLI labels in `p2m/cli.py:602-705` still use older terminology ("Suite", "Risk", "Sub-Risks", "Prompt Seeds") instead of the customer-facing terms (Dataset, Failure Mode, etc.). Worth aligning in a follow-up.
- The two pre-existing `tests/test_runner_progress.py` failures should be triaged separately.


jakepresent (commented):
Reviewed end-to-end. Substance is solid - the README rewrite, the docs map, the concept normalization, and the examples cleanup all read well, and the deliberate exclusion list is the right call. Three things flagged below.

## Blocker: PR description says it ships an OTel concurrency fix that is not in the diff

The description claims:

> **OTel concurrency fix** - switch `LiveOTelExporter` cross-session lock from `threading.Lock` to `asyncio.Lock` to avoid event-loop deadlock when `concurrency > 1`.

But there are zero changes to `p2m/core/otel.py` or `p2m/core/otel_session.py` in the diff (verified with `git diff origin/main..audit/private-preview-readiness -- p2m/` returning empty).

This matters because Yeming flagged this exact bug as a P0 in today's standup (`threading.Lock` in async code at `otel_session.py:150` blocks event loop when concurrency > 1). If the fix was dropped from a previous version of this PR, the description should be corrected. If the fix is intended but missed, it should land before merge.

Either way, can you confirm which it is? I do not want this PR to ship under a description that claims a P0 fix that is not actually here.

## Nit: Validation section references tests that should pass on this branch but the regression-CI claim is weaker than it looks

You called out:

> Two `tests/test_runner_progress.py` failures are pre-existing on `origin/main`

Worth a sentence in the PR body or a follow-up issue so the next reviewer does not have to re-derive that they are pre-existing. Otherwise the green CI here gives a false sense of "all tests pass" when in reality two are skipped because they were already broken.

## Lower-priority observations (do not block)

- `docs/private-preview/access-and-repo-workflow.md` is excellent reference material but reads as somewhat internal. The "What stays internal" list explicitly names "engineering design debates", "LT review material", "TAM and OSS escalation docs", "private customer notes". On a repo that flips public on June 1, that file could be misread by an external collaborator as a list of stuff that *was* in the repo and got removed. Worth a one-line preface clarifying it is forward-looking maintainer guidance, not a forensic record of past contents.
- The `health_assistant_external.yaml` rephrase is the right call. Good catch.
- The `travel_planner_eval` concept consolidation across the 7 phoenix configs is clean. The judge rubric collapse from "harmful_travel_advice" to "quality + safety failure" is a meaningful semantic change but it is well-documented in the diff and matches what Sue mentioned about needing to support both modes.
- `behavior_count` reduction from 10/15 to 6 across configs: not commented on in the description. If this is intentional cost-tightening for preview, worth a sentence; if it is just normalization, also worth a sentence. Future maintainers will wonder why.

## Net read

Direction is right. The OTel fix discrepancy is the only thing I want resolved before approving - everything else is comment-grade feedback that can either land here or as follow-ups.

Posting as a non-blocking comment review since I am the Verification DRI not a code-quality reviewer; treat the OTel point as the one that needs a reply.

---

_(edit: backticks were getting double-escaped in the original post; re-rendering the same content with proper code formatting)_

changliu2:
@jakepresent Thanks - you were exactly right, I was trying to make this a docs-only PR, but then actually testing the CLI commands in the docs breaks with the concurrency bug, so I appreciate the catch. The OTel concurrency fix was promised in the description but had been dropped from the diff. **Just landed it as `dfadadd`** so the PR now matches the description.

## What was actually broken

`LiveOTelExporter._lock` was a `threading.Lock`, acquired around the whole clear-invoke-export cycle in `OTelTracedSession.run_turn`:

``python
lock.acquire()
try:
    self._live_exporter.clear()
    response_text = await invoke_callable(self._callable, ...)  # awaits inside the sync lock
    turn_spans = self._exporter.export_session(turn_id)
finally:
    lock.release()
``

The rollout stage runs every session on one event loop (`asyncio.run(rollout.execute(...))`), so when `rollout.concurrency > 1` the second coroutine could not acquire the threading lock because the first was awaiting an HTTP call on the same loop, and the first could never resume because the loop thread was parked on the sync lock. Hard deadlock.

I confirmed the bug locally first by trying to run the LangGraph travel-planner config with `rollout.concurrency: 4` and 20 seeds: heartbeat froze after the first batch, zero transcripts produced after 40+ minutes.

## The fix (`dfadadd`)

Two-part change:

**1. `p2m/core/otel.py` - per-loop async lock**

``python
class LiveOTelExporter:
    _lock: asyncio.Lock | None = None
    _lock_loop: asyncio.AbstractEventLoop | None = None

    @classmethod
    def get_lock(cls) -> asyncio.Lock:
        loop = asyncio.get_running_loop()
        if cls._lock is None or cls._lock_loop is not loop:
            cls._lock = asyncio.Lock()
            cls._lock_loop = loop
        return cls._lock
``

A class-level `asyncio.Lock()` would bind to whichever loop happened to be running at first use, then raise on any subsequent `asyncio.run` (relevant for tests and repeated invocations). Lazy per-loop creation handles both the production case (one rollout = one loop, all sessions share the lock) and the multi-loop case (each `asyncio.run` gets a fresh lock).

**2. `p2m/core/otel_session.py` - `async with`**

``python
lock_ctx = LiveOTelExporter.get_lock() if self._live_otel else nullcontext()
async with lock_ctx:
    ...  # clear -> await invoke_callable -> export
``

`async with` on an `asyncio.Lock` yields the loop while waiting, so concurrent sessions interleave correctly. `contextlib.nullcontext()` is used in the non-live-OTel path so the rest of the body stays unchanged.

## Verification

- `uv run pytest -q` -> **505 passed, 14 skipped** (full suite, including all `OTelTracedSession` tests).
- LangGraph travel-planner with `rollout.concurrency: 4`, 5 behaviors, 10 prompts + 10 scenarios end-to-end (20 seeds total): **PASS in 375.3s, 20 transcripts, 20 scores**. No deadlock, heartbeat updated continuously.

Bulk verification across all customer-facing configs at `concurrency: 4` is in flight - I'll post a follow-up comment with the per-config table once it lands.

## Other points from your review

- **Pre-existing test failures**: I'll add a sentence to the PR description noting that `test_runner_progress.py` failures predate this PR. (Also fixed an orthogonal stale assertion in `test_systematization_stage.py` in `e2e9164` while I was in the test tree - it expected an obsolete prompt heading and didn't match what main has.)
- **`access-and-repo-workflow.md` framing**: agreed, will add a one-line preface clarifying it's forward-looking maintainer guidance.
- **`behavior_count` reduction (10/15 -> 6)**: intentional cost-tightening for preview - will document in the PR body.

Treating the OTel point as resolved; the other items are non-blocking polish I'll fold in the next push.

jakepresent (approved):
Approving. The OTel fix in dfadadd is correct and matches what Yeming flagged as the P0 in standup - per-loop asyncio.Lock cached lazily is the right shape (class-level asyncio.Lock() would have bound to the first loop and broken on repeat asyncio.run calls, and the inline comment in otel.py calls that out for future maintainers). The async with in otel_session.py plus nullcontext() for the non-live-OTel path keeps the rest of the body unchanged, which is what I want for a fix this load-bearing.

Verification is solid: full suite 505 pass / 14 skip, and the concurrency: 4 LangGraph travel-planner repro now completes 20/20 transcripts in ~375s where pre-fix it deadlocked indefinitely. Tier 1 + Tier 4 + CLA all green.

Non-blocking polish for whenever you do the next push:
- bump the test count in the validation block (502 → 505) and explicitly note test_runner_progress.py is pre-existing on origin/main
- one sentence on the behavior_count 10→6 rationale being intentional cost-tightening for preview

Treating those as ship-and-fold rather than gating. Nice catch on the health_assistant_external rephrase too.

PR #24, opened 2026-05-06
fix: improve error handling in policy generation and rollout

tangym:
## Summary

Fix two error handling issues that caused noisy failures and poor user experience during pipeline runs.

### Commit 1: Retry structured policy generation on transient parse failures

**Problem:** The policy stage calls `generate_structured()` which can return a valid HTTP response with malformed JSON that doesn't parse into a policy dict. This immediately raised `ValueError` and failed the pipeline, even though a simple retry usually succeeds.

**Fix:**
- Retry structured generation up to 2 attempts on parse failure
- Log a warning on each retry attempt
- Include actionable hint in error message when retries exhausted: "rerun the command to retry" and "check your endpoint's token rate limit and quota"
- Include last response text (truncated) for debugging

Note: The rate limiter (`_with_retries` inside `generate_structured()`) handles HTTP-level errors (429/5xx). This retry is a separate layer for content-quality failures where the HTTP call succeeded but the output is unusable.

### Commit 2: Propagate LLM errors from rollout instead of burying in transcript

**Problem:** Two broad `except Exception` blocks in `rollout.py` caught classified LLM errors (`LLMRateLimitError`, `LLMAuthError`, etc.) and either:
- Treated rate-limit errors as "bad auditor output" and retried with a guidance prompt (auditor loop, L659)
- Buried LLM errors as `[TARGET ERROR]` in the transcript, then surfaced them as `RuntimeError: scenario ended with target_error` with a noisy traceback (target turn loop, L741)

**Fix:** Add explicit catch for `(LLMAuthError, LLMInputError, LLMRateLimitError, LLMProviderError)` before the broad `except Exception` that re-raises immediately. The runner's top-level handler at L393 already presents these with clean, actionable messages. Non-LLM errors (runtime crashes, JSON parse failures) still follow the existing transcript recording path.

## Testing

```
uv run pytest -q  # 528 passed (1 pre-existing docker permission failure)
```

### DIFF
--- a/p2m/stages/rollout.py
+++ b/p2m/stages/rollout.py
@@ -37,6 +37,7 @@ from p2m.core.io import (
     row_factors,
 )
 from p2m.core.model_client import GenerateOptions, Message, ModelResponse, build_llm_call_trace, generate, to_jsonable
+from p2m.core.model_client import LLMAuthError, LLMInputError, LLMRateLimitError, LLMProviderError
 from p2m.core.session import (
     CallableSession,
     ExternalSession,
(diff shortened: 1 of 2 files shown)


tangym:
@microsoft-github-policy-service agree company="Microsoft"

AaronAspinwall123 (approved):

PR #32, opened 2026-05-08
 Hashing and rerun logic

AaronAspinwall123:


jakepresent (commented):
Real piece of work, the versioned cache + hash chain is the right structure. Two concerns I'd want addressed before this lands:

1. The Copilot findings on `_metadata_outputs_exist` and `_metadata_output_paths` are legitimate. `metadata['files']` should be validated as basenames before use, and `activate_latest_artifacts` should fall back when the primary output key is missing from metadata - both are real failure modes, not theoretical.
2. If a stage fails after `prepare_artifact_plan` has allocated `v####` but before `finalize_artifact_plan` runs, we leave a partial version directory with no sidecar. Subsequent runs handle this fine (the sidecar absence skips it), but it's a disk leak. Worth either cleaning up on failure or adding a `p2m results gc` later.

Two non-blocking observations:

- The concept hash is only computed for the policy stage; downstream stages pick it up transitively through the dependency chain. That's correct but assumes dependency chain coverage. A comment in `_stage_descriptor` about why this is safe would help.
- `override_cacheable_output_paths` silently overrides `save_dir` / `save_path` from user YAML. Should log when overriding so customers don't get confused why their `save_dir: ~/runs/foo` is ignored.

Tests look solid. Approving once the Copilot findings are addressed.

changliu2:
## Audit (PR-merge readiness)

Pulled `pr-32` locally and ran the full suite + diff review against `origin/main` (merge-base `da841a8`, merges cleanly). Net: **design is solid, recommend approve with one ask before merge**.

**Tests:** 73 new artifact-cache tests pass; 578 passed / 14 skipped on the full suite. The 2 `test_logging_config.py` failures are pre-existing on `main` (Windows tempfile cleanup, introduced by #22), not from this PR.

**False alarm I want to flag so reviewers don't chase it:** the `rollout.py:1018` change from `"\u2713"` to `"✓"` looks scary in `git diff` (mojibake on cp1252 terminals) but the on-disk bytes are clean UTF-8 (`E2 9C 93`). Same Python string, just a literal-vs-escape stylistic change. Worth a squash-on-merge so it doesn't confuse future reviewers — see (5) below.

### P1 — please address before merge

1. **Race in `_next_version` (`p2m/core/artifact_cache.py:675`).** Two concurrent `p2m run` invocations on the same suite both read `max(numbers) + 1`, both pick `v0001`, second clobbers first. No file lock or `mkdir(exist_ok=False)` retry. Single-process is fine, but anyone running parallel pipelines on a shared suite (e.g. CI matrix, batch sweeps) will silently corrupt each other's artifacts and `latest.json`. Cheap fix: wrap the allocation in a small loop that does `mkdir(exist_ok=False)` and bumps the version on `FileExistsError`. Alternatively, document "single-process per suite" as a constraint.

2. **`refresh_compatibility_files` silently overwrites suite-root files (`artifact_cache.py:371`).** `shutil.copy2(path, suite_root / path.name)` runs unconditionally on every reuse and finalize. A user who hand-edits `<suite>/policy.json` between runs has it silently destroyed the next time they hit a cache. At minimum, hash-check the destination before copy and warn if it diverges from the cached version (which is the user's signal that they've edited it).

### P2 — minor, can be follow-ups

3. **`_PROMPT_FILES` is hardcoded (`artifact_cache.py:83`).** Only the named default templates are hashed. If a stage config points at a custom prompt file, the cache won't invalidate when that file changes. Either scan the stage config for `*_path` keys ending in `.md` and include their hashes, or add a doc note.

4. **`_record_run_artifacts` runs 3+ times per pipeline** (`runner.py:438, 481, 517`). Idempotent and tiny, but wasteful. Could move to a single call at the end of `run_pipeline` since `artifact_versions` is fully resolved before any run-scope stage starts.

5. **Squash on merge.** History has 10 commits including 4 "Address Copilot review round N" + 2 "Potential fix for pull request finding" + the cosmetic `\u2713`→`✓` change mixed in. A squash would give us one clean entry against `main` and remove the false-alarm-bait.

### Things I checked that look correct

- `_metadata_outputs_exist` requires both `policy.json` AND `systematization.json` for a policy cache hit. Verified `run_systematization` writes its file unconditionally, so this won't permanently miss.
- `manifest is None` guard at the end of `run_pipeline` (`runner.py:514`) is in place — the new final `_record_run_artifacts` call is reachable only when `manifest` is non-None.
- Viewer `listRunIds` correctly filters out the new `artifacts/` directory so it doesn't show up as a phantom run.
- `rewrite_seed_path`'s two-layer protection (the boolean from absent `artifact_versions["seeds"]` PLUS the `_is_versioned_seed_artifact_path` check inside `run_rollout`) prevents clobbering versioned cache files. Backwards-compatible with users who supply their own seed file.
- Path-traversal defense is layered in three places (`_resolve_ref_path`, `_seed_artifact_path` in `viewer_read_model.py`, `manifestArtifactPath` in `viewer/src/lib/server/artifacts.ts`) — relative `..` rejected, absolute paths must resolve inside suite. Good.

### Branch state

PR is rebased/merged with `main` via the merge commit `ee30386` and `git merge-tree --write-tree` exits clean — GitHub merge will succeed. No action needed there.

changliu2:
## Re-audit (post `d12e7de` + `ba31928`)

Pulled the latest two commits, re-ran tests, and verified both P1 fixes against the round-1 audit.

**Tests:** 594 passed / 14 skipped (+16 new tests since round 1; the 2 deselected `test_logging_config.py` Windows tempdir failures are pre-existing on `main` from #22, unchanged here).

### P1 (1) Race in `_next_version` → ✅ fixed

Replaced with `_allocate_version_dir(stage_root)`:
- Loops up to `_MAX_VERSION_ALLOCATION_RETRIES=100`, calls `mkdir(parents=False, exist_ok=False)` on the candidate.
- On `FileExistsError`, rescans and retries (handles both "concurrent allocator beat us" and "leftover empty dir from a crashed run").
- On exhaustion, raises `RuntimeError` with diagnostic context (loud failure, not silent misnumbering).
- Docstring is honest about the remaining last-writer-wins gaps in `update_latest` and `refresh_compatibility_files`, and recommends "one suite directory per concurrent pipeline" as the strong-isolation answer. That's the right scope.

Also tightens `discard_artifact_plan`'s safety: each non-reused plan now provably owns its `artifact_dir` (reserved by atomic mkdir), so failure-path `rmtree` can never wipe a sibling process's work.

Test coverage (`AllocateVersionDirTest`):
- Empty stage_root → v0001
- Existing v0001/v0002 → v0003
- **Race injection**: monkeypatched `_iter_version_dirs` reports empty on the first scan and pre-creates `v0001` mid-call so our `mkdir` raises `FileExistsError`; loop rescans and picks `v0002`. (`test_retries_when_candidate_dir_exists`)
- **Pathological collisions**: monkeypatched `mkdir` to always `FileExistsError` → `RuntimeError`. (`test_raises_after_exhausting_retries`)

### P1 (2) `refresh_compatibility_files` silent clobber → ✅ fixed

New `_is_local_edit(suite_root, stage_name, dest, source)` runs before every overwrite. Conservative: returns False (safe to copy) when ANY of (a) dest missing, (b) dest not regular file, (c) `dest_hash == source_hash`, or (d) `_was_cached_artifact(dest_hash)` is True. Only when all four fail do we conclude "user hand-edited the suite-root copy", log a warning naming the file + cached source + `--force-stage` remediation, and skip.

The (d) check (walking every `vNNNN/artifact.json` for matching `file_hashes`) is what keeps `--force-stage <stage>` transparent: when a fresh `vNNNN` is produced and the suite-root copy still holds the prior version's content, the prior version's recorded hash recognizes it as cache-derived and the new content overwrites cleanly. Nice — that's the subtle case I would have asked about, and the doc and the test both call it out explicitly.

Test coverage (`RefreshCompatibilityFilesTest`):
- Copies when destination missing
- No-op when destination matches source (no spurious warning)
- **Preserves user edit and warns** — the actual P1 case
- **--force-stage path overwrites when destination matches prior cached version** — the transparency case
- Per-file isolation across multi-output stages

### Bonus fixes from `d12e7de` (Jake's review)

3. **Disk-leak on stage failure** → `discard_artifact_plan` wired into the runner exception path. No-ops for reused plans (correctly: a downstream failure must not blow away a healthy upstream cache hit). `latest.json` left alone (correctly: `finalize_artifact_plan` is its only writer for non-reused plans, so a discarded plan never touched it). Tests cover missing-dir, reused-plan no-op, ctx cleanup, and version-slot reuse after discard.

4. **Silent override of user `save_dir`** → `override_cacheable_output_paths` now `log.warning`s per overridden key with stage, key, user value, cache location, and `--force-stage` remediation. Skipped when no user value was set. Tests cover both branches.

5. **Concept-hash transitive flow** → comment in `_stage_descriptor` documents that `concept_hash` is computed only for `policy` and propagates to design/seeds via `_dependency_descriptor`, with the safety invariant ("every cacheable stage must depend on its upstream") spelled out so a future cacheable stage that breaks the chain trips a code-review flag rather than silently reusing stale outputs after a concept edit.

### Carried forward from round 1 (P2, fine as follow-ups)

- `_PROMPT_FILES` is still hardcoded — custom prompt paths supplied via stage config won't invalidate the cache. Documentation note would suffice.
- `_record_run_artifacts` still writes 3+ times per pipeline. Idempotent and tiny; not worth blocking.
- History is now 12 commits including 4 "Address Copilot review round N" + 2 "Potential fix for pull request finding" + the cosmetic `\u2713`→`✓` change. **Still recommend squash on merge** — gives `main` one clean entry and removes the false-alarm-bait diff.

### Verdict

P1 issues from round 1 are fully addressed with code, comments, AND focused regression tests. **LGTM, ship it.** Squash-on-merge if you can.

changliu2:
## Round 3 — re-audit after `main` advanced (PR #27 + PR #26 merged)

PR head unchanged at `ba31928`. Verified the would-be merged state still ships clean:

- **Merge:** `git merge-tree --write-tree origin/main pr-32` exit 0. Local test-merge of `pr-32` into `origin/main` (`26dd5af`) succeeds with no conflicts.
- **Tests on merged tree:** `594 passed, 14 skipped, 2 deselected, 13 subtests passed in 42.93s` — identical to round 2. The 2 deselects are the pre-existing Windows logging flakes from PR #22 (also fail on `main`, unrelated).
- **Ruff:** 12 errors, all pre-existing on `main` — none introduced by this PR.
- **Pyright:** 6 errors in `artifact_cache.py` are typing-narrowing nits (e.g., `Optional[Path]` not narrowed through ternary branches; one dataclass `asdict` ClassVar quirk). Not runtime bugs. No CI gate on pyright today, so non-blocking — worth a follow-up cleanup if/when pyright gets enforced.

Verdict unchanged: **ship it** (squash-on-merge). Nothing in the recent main advance affects this PR's correctness.

p2m/core/artifact_cache.py
```diff
@@ -0,0 +1,1023 @@
...
+                    ctx,
+                    stage_name=stage_name,
+                    version=version,
+                    artifact_dir=artifact_dir,
+                    metadata=metadata,
+                    primary_path=output_paths[next(iter(_OUTPUT_FILES[stage_name]))],
+                )
+                update_latest(ctx, stage_name, ref)
+                print(
+                    f"[artifact-cache] warning: latest.json {stage_name} entry "
```
tangym, line 317: Nit: change `print(file=sys.stderr)` to `log.warning()`

PR #22 moved all warning/error output to `log = logging.getLogger(__name__)` so `--verbose/--quiet/--log-file/--output json` work uniformly. This module already uses `log.warning()` in discard_artifact_plan and refresh_compatibility_files, but several other sites still use `print(..., file=sys.stderr)` which bypasses log level filtering, `--quiet`, and JSON output mode. Switching these to log.warning() with structured placeholders (like the existing calls) would keep the module consistent with the rest of the codebase.

tangym (approved):

PR #44, opened 2026-05-12
rollout/judge: per-seed error isolation with batch-level error-rate threshold

jakepresent:
## Problem

A single content-filter / input-refusal hit was killing entire scale runs. The May 11 overnight scale-500 was the first one: seed 501 of 502 was rejected by Azure CF, and 500 already-completed transcripts were thrown away. As subsequent runs ran further, three more independent places surfaced the same failure mode:

- **callable-target raw `litellm` exceptions** - LangGraph (and any framework wrapper) makes provider calls inside the user callable, so `litellm.BadRequestError` etc. bypass `generate()`/`_with_retries` entirely and the rollout's `isinstance(_, LLMInputError)` checks miss them.
- **auditor-side input refusals** - the adversarial prompt the auditor sends to the target trips Azure Prompt Shields' jailbreak detector. Surfaced live at seed 853/1000 on the May 12 mix run.
- **judge-side input refusals** - the same adversarial content the auditor produced trips the content filter on the judge call itself. Surfaced at scored-seed 539/1000 on the May 12 rerun.
- **non-classified runtime errors inside the agent under test** - `openai.BadRequestError` raised inside a LangGraph internal task ('research') escaped past both `_run_prompt_seed` and `CallableSession`'s wrapper because it was raised in a different `asyncio` task than the one being awaited, so it never reached the classification layer at all.

This is the wrong failure mode for an adversarial-eval pipeline. Generating prompts the target / auditor / judge refuses is exactly what we're supposed to do, and the refusal itself is data we want to keep, not a global pipeline error.

## Fix

Treat per-seed failures as **data**, not as global pipeline errors, at four layers, plus a generic catch-all with a batch-level error-rate threshold.

### Per-layer isolation

| Layer | What's caught | Recorded as |
|---|---|---|
| **Target side** (`rollout._run_prompt_seed`, `_run_auditor_target_loop`) | `LLMInputError` from the target call | `[TARGET INPUT REFUSED: ...]` event, `stop_reason='target_input_refused'` |
| **Callable target classification** (`session.CallableSession.run_turn`) | Raw provider exceptions raised inside user callables | Wrapped via `_classify_llm_error` so they flow through the same target-side isolation; unclassified errors propagate unchanged |
| **Auditor side** (`rollout._run_auditor_target_loop`) | `LLMInputError` from the auditor call | `[AUDITOR INPUT REFUSED: ...]` event, `stop_reason='auditor_input_refused'` |
| **Judge side** (`stages.judge`) | `LLMInputError` on the judge call | Synthetic score row with `judge_status='filter_skipped'`, `judge_error='judge_input_refused: ...'`; flows through the existing `infer_judge_status()` contract as a non-ok seed |

### Generic catch-all (final commit)

Any non-global exception that escapes the targeted handlers now synthesizes a transcript with `stop_reason='runtime_error'` carrying a single system event with the error type and message. The transcript is written like a normal one; the seed isn't lost. The exception is still attached to the result so the gather loop can count it.

### Batch-level error-ratio threshold

Replaced `if errors: raise errors[0]` in the rollout gather loop with a configurable threshold:

- Default **10%** (1 in 10 seeds erroring is the tipping point)
- Tunable via `P2M_ROLLOUT_ERROR_FAIL_RATIO` for ops scenarios
- Below threshold: log a single warning summary with the ratio, stage completes normally
- Above threshold: log an error and raise the first exception so the runner surfaces a clean message

Systemic failures (auth misconfigured, every seed errors) still fail fast.

### What still fails fast (intentionally)

- `LLMAuthError` - global config problem; retry won't help
- `LLMRateLimitError` - global throttling
- `LLMProviderError` - provider 5xx; conservative choice
- Any pre-rollout stage failure (policy/design/seeds) - those have no per-seed semantics

## Validation

End-to-end 1k mix run on `15332c8` (`travel-planner-langgraph-pr44-mix1k-e2e/baseline`, 700 prompts + 300 scenarios, C=5):

| Stage | Wall | Per case |
|---|---|---|
| policy | 16.2s | - |
| design | 3.1s | - |
| seeds | 72.9s | - |
| rollout | 1796.7s (29.9 min) | 1.80s |
| judge | 1925.8s (32.1 min) | 1.93s |
| **total** | **3814.7s (1h 3m 36s)** | |

1000/1000 transcripts, 1000/1000 scored, all `judge_status='ok'`, zero recorded runtime errors. Earlier runs on the same branch did exercise the isolation paths live (auditor refusal at 853, judge filter at 539, LangGraph escape ~994); the final run was clean.

## Tests

7 new tests:

- `test_run_turn_reclassifies_litellm_bad_request_as_input_error` (session)
- `test_run_turn_passes_through_unclassified_exceptions` (session)
- `test_run_prompt_seed_records_target_input_refusal` (rollout)
- `test_run_rollout_isolates_target_input_refusal_to_one_seed` (rollout)
- `test_run_rollout_isolates_auditor_input_refusal_to_one_seed` (rollout)
- `test_run_rollout_still_fails_fast_on_provider_5xx` (rollout)
- `test_run_rollout_tolerates_below_threshold_errors` (rollout)

1 existing test updated (`test_run_rollout_keeps_partial_successful_transcripts_when_later_worker_fails`) to reflect the new contract: failing seeds now have synthesized `stop_reason='runtime_error'` transcripts, and the 50% error rate still aborts.

All 52 rollout + exception-handling tests pass locally.

## Files

```
p2m/core/session.py              |  37 +++-
p2m/stages/judge.py              |  36 +++-
p2m/stages/rollout.py            | 194 ++++++++++++++++---
tests/test_exception_handling.py |  69 +++++++
tests/test_rollout_stage.py      | 401 +++++++++++++++++++++++++++++++++++-
5 files changed, 702 insertions(+), 35 deletions(-)
```

## Commits (read in order)

Each commit closed a real gap surfaced by the previous commit's validation, so the sequence is also the discovery log:

1. `82cf339` - target-side `LLMInputError` isolation (the original May 11 fix).
2. `0184d8d` - classify raw `litellm` exceptions raised inside user callables, so callable-target frameworks (LangGraph, etc.) flow through the same target-side isolation.
3. `f265154` - auditor-side `LLMInputError` isolation.
4. `dcaa91f` - judge-side `LLMInputError` isolation.
5. `15332c8` - generic per-seed error isolation with batch-level error-ratio threshold.

## Known follow-ups not in this PR

- **Seeds-stage content-filter refusals are not isolated.** Generation is pre-rollout, so the per-seed semantics here don't apply directly. Separate work.
- **10% default error-ratio threshold is a starting value, not a calibrated one.** For 10-seed runs it's 1-error sensitive; for 1k+ runs it allows up to 100 silent errors. Worth revisiting once we have multi-run error-rate data.
- **`P2M_ROLLOUT_ERROR_FAIL_RATIO` env var is a temporary control surface.** If it stays useful it should graduate to a real config field; if not, it can be dropped.
- **Overlaps with Aaron's PR #45 benchmark shim** (`_install_content_filter_tolerance` in `scripts/benchmark.py`, using `stop_reason="content_filter_blocked"`). That patch monkey-patches the rollout to drop content-filter-blocked seeds at benchmark time; this PR lands per-seed isolation in product code so the shim can be retired once #45 merges (or revisited to align stop_reason naming).

### DIFF
--- a/p2m/core/session.py
+++ b/p2m/core/session.py
@@ -15,6 +15,7 @@ from p2m.core.model_client import (
     Message,
     ModelResponse,
     ToolCall,
+    _classify_llm_error,
     build_llm_call_trace,
     generate,
     generate_with_tools,
@@ -512,15 +513,35 @@ class CallableSession:
                 for msg in messages
                 if msg.role in ("user", "assistant")
             ]
-            raw_result = await invoke_callable(
-                self._callable, user_text, history=history,
-                timeout_s=self._message_timeout_s,
-            )
+            try:
+                raw_result = await invoke_callable(
+                    self._callable, user_text, history=history,
+                    timeout_s=self._message_timeout_s,
+                )
+            except Exception as exc:
+                # The user callable typically makes its own LLM calls
+                # (LangGraph, agent frameworks, raw litellm) which bypass
+                # our generate()/_with_retries wrapper, so provider errors
+                # bubble up unclassified. Re-raise them as the right p2m
+                # error class (LLMInputError for content-filter / 400s,
+                # LLMRateLimitError for 429s, LLMProviderError for 5xx,
+                # LLMAuthError for 401/403) so the rollout stage's per-seed
+                # isolation paths can do their job.
+                classified = _classify_llm_error(exc)
+                if classified is exc:
+                    raise
+                raise classified from exc
         else:
-            raw_result = await invoke_callable(
-                self._callable, user_text,
-                timeout_s=self._message_timeout_s,
-            )
+            try:
+                raw_result = await invoke_callable(
+                    self._callable, user_text,
+                    timeout_s=self._message_timeout_s,
+                )
+            except Exception as exc:
+                classified = _classify_llm_error(exc)
+                if classified is exc:
+                    raise
+                raise classified from exc
 
         result = self._normalize_callable_result(raw_result)
 
--- a/p2m/stages/judge.py
+++ b/p2m/stages/judge.py
@@ -170,7 +170,41 @@ async def run_judge(
                 "output_index": output_index,
                 "score_row": await score_row(row),
             }
-        except (LLMAuthError, LLMInputError, LLMRateLimitError, LLMProviderError):
+        except LLMInputError as exc:
+            # Judge-side input refusal (e.g. Azure content filter rejecting
+            # a transcript whose adversarial content the judge LLM can't
+            # process). This is per-seed data, not a global pipeline
+            # problem: a different transcript will judge cleanly. Record a
+            # filter_skipped score row so the seed isn't lost and the
+            # stage can move on. Mirrors the target_input_refused and
+            # auditor_input_refused handling in rollout.
+            seed_id = row.get("seed_id", "?")
+            log.warning(
+                "Judge content-filter refusal for seed %s: %s",
+                seed_id, exc,
+            )
+            factors = row_factors(row)
+            score_row_filter_skipped = {
+                "kind": row.get("kind", ""),
+                "seed_id": seed_id,
+                "concept": row.get("concept", ""),
+                "judge_model": judge_model,
+                "target": row.get("target", ""),
+                "auditor_model": row.get("auditor_model", ""),
+                "judge_status": "filter_skipped",
+                "judge_error": f"judge_input_refused: {exc}",
+                "verdict": {},
+            }
+            if factors:
+                score_row_filter_skipped["factors"] = factors
+            return {
+                "output_index": output_index,
+                "score_row": score_row_filter_skipped,
+            }
+        except (LLMAuthError, LLMRateLimitError, LLMProviderError):
+            # Auth/rate-limit/provider-5xx errors are global pipeline
+            # problems, not seed-specific. Propagate so the runner can
+            # surface a clean message and fail the stage fast.
             raise
         except (json.JSONDecodeError, ValueError) as exc:
             seed_id = row.get("seed_id", "?")
--- a/p2m/stages/rollout.py
+++ b/p2m/stages/rollout.py
@@ -7,6 +7,7 @@ from dataclasses import asdict
 import hashlib
 import json as json_module
 import logging
+import os
 import re
 import traceback
 import uuid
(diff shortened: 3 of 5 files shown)


jakepresent:
Pushed [0184d8d](https://github.com/microsoft/adaptive-eval/commit/0184d8d): `CallableSession.run_turn` now classifies exceptions escaping `invoke_callable` through `_classify_llm_error` before they propagate.

**Why the second commit:** the rollout isolation in the first commit only fires when the exception reaches the helpers as an already-classified `LLMInputError`. For `callable:` targets (LangGraph agents, framework wrappers, raw litellm callers), the user's own code makes the provider call and bypasses `generate()`/`_with_retries` entirely, so provider errors bubble up as raw `litellm.BadRequestError` and the `isinstance(_, LLMInputError)` checks miss them. Same failure pattern the isolation was meant to fix. Caught this locally when smoke-testing the PR against the LangGraph travel-planner callable.

Classified errors are re-raised with `__cause__` preserved; unclassified exceptions (user agent crashes, etc.) pass through unchanged so they don't get smuggled into one of the four LLM error classes.

Two new tests in `tests/test_exception_handling.py`:
- `test_run_turn_reclassifies_litellm_bad_request_as_input_error`
- `test_run_turn_passes_through_unclassified_exceptions`

50/50 rollout + exception-handling tests pass.

jakepresent:
Pushed [f265154](https://github.com/microsoft/adaptive-eval/commit/f265154): auditor-side LLMInputError now gets the same per-seed isolation as the target-side case.

Hit this live on the mix-1k validation run: at scenario 853/1000 the auditor's adversarial prompt for one seed tripped Azure Prompt Shields' jailbreak detector. Same Bad request / 'flagged as potentially violating our usage policy' signature, just on the auditor side instead of the target side. This is the same pattern Aaron flagged in standup today and worked around in PR #45 with a tame auditor prompt.

The previous two commits handled the target side and the callable-target classification gap. This one closes the loop: `_run_auditor_target_loop` now catches `LLMInputError` separately from the other classified errors, records an `[AUDITOR INPUT REFUSED: ...]` system event in the transcript with `stop_reason='auditor_input_refused'`, and ends the conversation cleanly. Auth/rate-limit/5xx errors still propagate.

One new test, `test_run_rollout_isolates_auditor_input_refusal_to_one_seed`, reproduces the May 12 scenario at 5-seed scale: one seed's auditor call raises `LLMInputError`, the other four complete cleanly, the batch returns successfully. 51/51 rollout + exception-handling tests pass.

Relaunching the 1k mix overnight on this commit.

jakepresent:
Pushed [dcaa91f](https://github.com/microsoft/adaptive-eval/commit/dcaa91f): judge-side LLMInputError now gets the same per-seed isolation as the rollout-side cases.

Hit this live on the mix-1k rerun. After the auditor-side fix from f265154 unblocked rollout, the judge stage made it through 539 of 1000 transcripts (at C=3 to avoid the rate-limit cliff at C=10) before one transcript's adversarial content tripped Azure's content filter on the judge call itself. Same Bad request / 'flagged as potentially violating our usage policy' signature, just on the judge side instead of target or auditor.

This is the third logically-independent place the same adversarial-eval-as-attacker dynamic fires, so PR #44 has grown into a coherent 'isolate LLMInputError per-seed across all three stages' shape:
- target-side (`82cf339`): rollout records `stop_reason=target_input_refused`
- callable target classification gap (`0184d8d`): `CallableSession.run_turn` classifies raw litellm exceptions so the rollout fix actually fires on `callable:` targets
- auditor-side (`f265154`): rollout records `stop_reason=auditor_input_refused`
- **judge-side (`dcaa91f`):** judge synthesizes a `judge_status=filter_skipped` score row with `judge_error=judge_input_refused: ...`

The `filter_skipped` status flows cleanly through `infer_judge_status()` in `p2m/core/judge.py` as a non-ok status, so `results.py` and `viewer_read_model.py` correctly count these as non-scored seeds without any further changes. Matches Aaron's PR #45 `_install_judge_skip_blocked` shim behavior, but lands it in product code instead of as a benchmark monkey-patch.

Re-running the mix-1k now on the new HEAD.

jakepresent:
Pushed [15332c8](https://github.com/microsoft/adaptive-eval/commit/15332c8): generic per-seed error isolation with a batch-level error-ratio threshold (default 10%).

The mix-1k validation tonight surfaced a fourth specific failure mode that the targeted classification fixes don't catch: an `openai.BadRequestError` raised inside a LangGraph internal task escaped past `_run_prompt_seed` because it was raised in a different asyncio task than the one being awaited. Classification only catches errors that reach the classifier.

Rather than chase the next specific gap, this commit adopts a generic guarantee: **one bad conversation can't kill the whole run.**

- The worker now catches all non-global exceptions and synthesizes a transcript with `stop_reason='runtime_error'` carrying the error message. The seed is preserved as data, not lost.
- The gather loop replaces `if errors: raise errors[0]` with an error-ratio threshold (default 10%, tunable via `P2M_ROLLOUT_ERROR_FAIL_RATIO`). Below threshold: warn and continue. Above threshold: abort the stage. Systemic problems (deployment misconfigured, every seed errors) still fail fast.
- Auth/rate-limit/provider-5xx still propagate as before (these are global, not seed-specific).

PR #44's shape has now stabilized into 'isolate per-seed failures, fail the stage only on systemic problems.' Five commits total:
- `82cf339` target-side LLMInputError
- `0184d8d` callable target classification gap
- `f265154` auditor-side LLMInputError
- `dcaa91f` judge-side LLMInputError
- **`15332c8` generic per-seed error isolation + ratio threshold**

52/52 rollout + exception-handling tests pass.

jakepresent:
Closing as superseded by #45. Aaron's PR absorbed the per-row/per-batch error-isolation pieces from this branch and now has the cache-finalization and benchmark-summary fixes on top.

PR #45, opened 2026-05-12
Load testing

AaronAspinwall123:


jakepresent:
Review pass found two correctness issues I think we should fix before relying on this for load-test results.

1. Partial seed failures can be cached as successful artifacts

`seeds.py` now allows partial seed-generation success: `_generate_records()` returns successful records plus `errored_count` when some batches fail, and `run_seeds()` writes only the successful records to `seeds.jsonl`. But the stage wrapper drops `errored_count`, and `runner.py` finalizes cacheable suite artifacts whenever the stage returns without raising. That means a transient partial seed failure can produce a smaller-than-requested `seeds.jsonl`, get an `artifact.json` sidecar, update `latest.json`, and then be reused later because the input hash still matches.

Relevant refs:
- `p2m/stages/seeds.py:776-793` tolerates partial batch failures
- `p2m/stages/seeds.py:858-876` writes partial records and returns `errored_count`
- `p2m/stages/seeds.py:1000-1007` drops `errored_count` from the stage result
- `p2m/runner.py:536-543` finalizes cacheable artifacts on any non-raising stage result
- `p2m/core/artifact_cache.py:367-395` writes the sidecar / updates `latest.json`

Suggested fix: either make non-zero seed `errored_count` fail the cacheable seeds stage, or teach the runner/artifact cache not to finalize a cacheable artifact when the stage reports partial failure. If partial seeds are intentionally allowed, the artifact metadata also needs expected vs actual counts so reuse can distinguish “complete artifact” from “best-effort partial artifact.”

2. Benchmark CSV outcome fields are wired to the wrong metrics schema

`benchmark.py` expects `metrics.json` to contain seed/scenario outcome metrics like `scenario_metrics`, `seed_metrics`, `policy_violation_true_rate`, and `overrefusal_true_rate`. But this PR’s runner writes `metrics.json` as token-usage telemetry only: `stages`, `per_model`, and `totals`. So the benchmark CSV fields for `scenario_seeds_generated`, `scenarios_scored`, `policy_violation_true_rate`, and `overrefusal_true_rate` will stay blank even after a successful run.

There’s a second schema mismatch too: the existing result-summary helpers compute scenario metrics from `scores.jsonl` with keys like `total`, `scored_total`, and dimension `rate`, not `count` / `true_rate`.

Relevant refs:
- `scripts/benchmark.py:94-97` declares outcome CSV columns
- `scripts/benchmark.py:299-344` reads those fields from `metrics.json`
- `p2m/runner.py:250-295` builds token-usage-only `metrics.json`
- `p2m/runner.py:593-598` writes that payload
- `p2m/results.py:83-123` shows the existing score-derived metrics shape

Suggested fix: have `benchmark.py` compute these from `seeds.jsonl` / `scores.jsonl` directly, or call `p2m.results.load_run_summary(run_dir)` and map `scenario_metrics.total`, `scenario_metrics.scored_total`, and `dimensions[*].rate` into the CSV. Don’t read them from the new token-usage `metrics.json`.

Related smaller issue: rollout/judge return `errored_count` internally, but the stage wrappers drop it and the benchmark CSV only scans typed refusals / target errors / judge filter skips. Untyped per-row errors below the failure threshold can produce a successful exit code without any durable CSV signal. That can probably be fixed as part of the same benchmark-summary path.

jakepresent (approved):
Thanks for the quick fixes. The two correctness concerns I raised look addressed to me:

- Partial artifacts no longer get cache sidecars when `_summary.errored_count > 0`, so a partial seed output should not be reused as a complete cached artifact.
- `scripts/benchmark.py` now reads outcome columns from the seed/score artifacts instead of the token-usage-only `metrics.json`.

Focused tests looked good locally (`tests/test_benchmark_summary.py` plus the partial-seeds cache regression test), and CI is green.

Two small non-blocking follow-ups:

- Clean up the `ΓÇö` mojibake in a few comments/tests.
- Confirm the desired behavior for partial-seeds runs that exit successfully but skip cache finalization: since the suite-root compatibility `seeds.jsonl` may not refresh, `scenario_seeds_generated` in the benchmark summary could be blank or stale unless the summary reads the active artifact/run manifest instead. I don't think this blocks this PR, but it is worth deciding explicitly.

jakepresent (approved):
Re-approving after the two follow-up commits. The mojibake cleanup is comment-only, and the benchmark summary change is the right direction: for partial-seeds runs where cache finalization is skipped, sourcing the count from this run's scores avoids stale suite-root seeds.jsonl. Targeted benchmark/artifact-cache tests passed locally.

Issue #50, opened 2026-05-15
Centralize provider/model quirks (e.g. GPT-5 temperature=1.0); make example YAMLs .env-driven instead of Azure-hardcoded


changliu2:
## Symptom
Running any example with the default `azure/gpt-5.4*` deployments fails at the API boundary because Azure's GPT-5 family only accepts `temperature=1.0`. The current guard in `p2m/core/judge.py`, `p2m/stages/policy.py`, `p2m/stages/systematization.py`, and `p2m/stages/systematization_convert.py` only strips `temperature` when `reasoning_effort` is explicitly set in the YAML — not based on model name. So a config with `azure/gpt-5.4-mini` + `temperature: 0.7` and no `reasoning_effort` sends a request Azure rejects.

## Two related smells

1. **Provider quirks are scattered across stages.** Four separate files duplicate the same `if reasoning_effort is not None: temperature = None` block. There is no single layer that owns "this provider/model accepts which knobs."
2. **Example YAMLs are Azure-centric.** Every `examples/**/eval_config.yaml` and every file in `examples/pipes/*` hard-codes `azure/gpt-5.4-mini` or `azure/gpt-5.4`. A user with only `OPENAI_API_KEY` or `ANTHROPIC_API_KEY` set in `.env` cannot run any example without editing YAML. Notably, the example `agent.py` files *already* read `os.environ.get("P2M_TARGET_MODEL", ...)` — so the agent side is `.env`-driven but the eval side is not.

## Proposed direction (open for discussion)

- Add a single `model_capabilities` layer (probably in `p2m/core/model_client.py`) that maps model-name patterns → which sampling knobs are valid, and silently drops/coerces unsupported ones (with a one-line warning). Remove the four scattered `reasoning_effort`-keyed branches.
- Make example YAMLs read model from `${P2M_EVAL_MODEL}` / `${P2M_TARGET_MODEL}` via env interpolation (LiteLLM-style `provider/name`), with sensible defaults. This also means examples work out-of-the-box for non-Azure users.
- Document in `.env.example` and the README which env vars drive which slots (target / auditor / judge / policy / seeds).

## Open question

[...]

PR #56, opened 2026-05-15
provider: gate web search support

jakepresent:
## Summary

Adds a narrow provider capability gate for the current `web_search` transport path.

Today `web_search=True` always builds an OpenAI Responses API payload with `web_search_preview`. Gemini provider smoke testing showed that path fails before useful generation when combined with structured output, even though direct LiteLLM calls and a tiny manual rollout/judge smoke work.

This PR makes that mismatch explicit instead of letting non-OpenAI providers fail deep in LiteLLM/provider code.

## Review focus

- This is intentionally a small guardrail, not a full Gemini support PR.
- Main behavior to sanity-check: OpenAI / Azure OpenAI keep the existing `web_search_preview` path, while Gemini/Anthropic/etc. get a clear "disable `web_search`" error before provider calls.
- Follow-up remains separate: simplify/fallback the Gemini scenario-seed schema path.

## Changes

- Add a small model-family capability check in `p2m.core.model_client`.
- Allow the current `web_search_preview` path only for OpenAI / Azure OpenAI model families.
- Raise a clear `ValueError` for other providers telling users to disable `web_search` for that stage.
- Add unit coverage for Gemini structured generation and another non-OpenAI provider.

## Not included

This does **not** claim full Gemini support. The remaining Gemini-generated-path issue is scenario-seed schema complexity; that should be handled separately via schema simplification / fallback generation.

Related to #50.

## Test evidence

```bash
uv run pytest tests/test_model_client.py -q
# 7 passed
```

GitHub checks are green: Tier 1 unit tests, Tier 4 regression gate, and CLA.

### DIFF
--- a/p2m/core/model_client.py
+++ b/p2m/core/model_client.py
@@ -175,6 +175,39 @@ def build_json_schema_text_format(
     }
 
 
+def _model_family(model: str) -> str:
+    """Return the provider/model-family prefix used for transport capability checks."""
+    normalized = (model or "").strip().lower()
+    if "/" in normalized:
+        return normalized.split("/", 1)[0]
+    if normalized.startswith(("gpt-", "o1", "o3", "o4")):
+        return "openai"
+    return normalized
+
+
+def _supports_web_search_preview(model: str) -> bool:
+    """Whether this model can use the Responses API web_search_preview tool.
+
+    The current implementation sends OpenAI Responses API payloads with
+    ``tools=[{"type": "web_search_preview"}]``. LiteLLM exposes that
+    path for OpenAI-compatible models, but provider smoke testing showed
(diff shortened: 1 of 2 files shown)


changliu2 (approved):
Approving. Narrow, well-shaped: replaces a deep LiteLLM-internal failure with a clear pre-flight `ValueError`, and the test pair (`gemini/gemini-2.5-flash` + `anthropic/claude-sonnet-4`) asserts LiteLLM is never called on the rejection path, so we don't accidentally regress to the "fails deep in provider code" behavior this PR is fixing.

One non-blocking nit: `_model_family()` is family-prefix only, so `azure/gemini-pro` would falsely pass the gate. Nobody ships that combo today so not worth blocking, but worth a TODO if we extend the gate later.

PR #59, opened 2026-05-18
Defend harness against user-target cleanup deadlocks

AaronAspinwall123:
Adds three layers of runtime safety so the harness never deadlocks on a user-supplied target that leaves background resources open (unclosed httpx.AsyncClient instances, OTel exporters, etc.). Motivation: the multi-agent LangGraph travel-planner target hung the pipeline twice at s=1000 c=50, freezing inside asyncio.run() teardown between rollout and judge while loop.shutdown_default_executor() waited forever on worker threads stuck in cleanup finalizers.

Layer 1 - Bounded stage teardown (run_stage_coro)
  Replaces asyncio.run(module.run(ctx, raw_cfg)) in the runner with a
  variant that owns its own default ThreadPoolExecutor. On exit, async-gen
  shutdown + executor.shutdown(wait=True, cancel_futures=True) + loop.close
  run inside a daemon thread with a 300s bounded join. If the join times
  out, the executor is detached from concurrent.futures.thread._threads_queues
  (so process exit isn't blocked at atexit either) and a clear user-facing
  warning is logged. Pipeline continues to the next stage.

Layer 2 - PipelineWatchdog
  Daemon thread that dumps every Python thread's stack via
  sys._current_frames() if no progress tick fires for >600s. Makes future
  hangs self-diagnosing in the bench log instead of requiring py-spy.
  Idempotent per idle period: one hang -> one dump.

Layer 4 - ManifestHeartbeat
  Daemon thread that rewrites manifest.json every 30s with a fresh
  heartbeat_at and an optional progress payload (e.g. {stage: rollout,
  completed: 423, total: 1000, errors: 2}). Replaces the previous behavior
  where manifest.heartbeat_at was only refreshed at stage boundaries -
  meaning a 90-min rollout left it stale for the entire duration.

Wiring:
  - RunManifest gains an optional progress dict[str, Any] field.
  - rollout.run_rollout and judge.run_judge accept an optional heartbeat= kwarg and push live progress on every scenario / score completion.
  - The runner injects ctx[_heartbeat] so stages can pick it up via ctx, and starts/stops the heartbeat + watchdog in a single try/finally around an extracted _run_stages_inner helper.
  - The naive travel-planner example is intentionally NOT fixed - it keeps exercising Layer 1 on every benchmark run.

Tests (tests/test_runtime_safety.py, 12 cases):
  - run_stage_coro returns within timeout even when a worker leaks a thread that refuses to die (the headline regression test for the LangGraph deadlock pattern).
  - run_stage_coro propagates coroutine exceptions and logs no warning on clean shutdowns.
  - ManifestHeartbeat writes periodically, merges set_progress kwargs, drops the payload on clear_progress, and survives write_fn errors without killing the daemon.
  - PipelineWatchdog dumps stacks after the idle threshold, only once per idle period, and stays quiet when tick() fires regularly. The heartbeat ticks the watchdog automatically when attached.


jakepresent:
Thanks for pushing this - the shape is useful, and the tests cover the in-process stage-return path well. I think there is one blocking issue before this can be considered a fix for the deadlock class, though.

`run_stage_coro()` returns after the bounded cleanup timeout, but the abandoned executor workers are still normal non-daemon Python threads. `_detach_executor_from_atexit()` removes them from `concurrent.futures.thread._threads_queues`, which skips the `concurrent.futures` atexit join, but Python still waits for live non-daemon threads during interpreter shutdown.

I reproduced this with a subprocess probe that starts a default-executor worker, lets `run_stage_coro(..., cleanup_timeout_s=0.2)` return, and then exits without releasing the worker. The cleanup warning is logged and the function returns, but the process still does not exit; `/usr/bin/timeout 3s ...` kills it with rc=124.

That means the PR currently solves “continue to the next stage in-process,” but not “abandoned worker threads cannot block shutdown / will be reaped on process exit,” which the PR body and warning currently claim. The headline test also misses this because it releases the worker in `finally`.

Could we add a subprocess regression test that leaks the worker and asserts the child process exits quickly, then adjust the implementation accordingly? If full process-isolation is out of scope, I think the PR should at least narrow the claim and not imply process-exit safety.

jakepresent (changes requested):
Thanks for the follow-up. I think the original blocker is fixed for the Python version CI is exercising, but I found a remaining version-specific hole before I can approve this.

On the PR branch, running the new subprocess regression test under CPython 3.14.3 still hangs at interpreter shutdown:

```bash
uv run pytest tests/test_runtime_safety.py::test_run_stage_coro_does_not_block_subprocess_exit_when_worker_leaked -q
```

Result: the subprocess prints both `RESULT=main-done` and `EXIT-SENTINEL`, then times out after 15s. The failure message is the test's own assertion:

```text
Subprocess hung at interpreter shutdown (timeout=15.0s).
This means the worker-thread detach is incomplete: threading._shutdown is still joining the leaked worker.
```

I think CI is missing this because `.github/workflows/regression.yml` is pinned to Python 3.11. I also checked the same targeted test under CPython 3.12.3 and it passes there.

The likely root cause is that on 3.14.3, `threading._shutdown_locks` and `threading._shutdown_locks_lock` are absent/`None`, so `_detach_executor_from_atexit()` only removes the worker from `concurrent.futures.thread._threads_queues`. That leaves the newer interpreter shutdown path still waiting on the leaked non-daemon thread.

Since `pyproject.toml` currently says `requires-python = ">=3.11"`, I think we either need the detach path to handle 3.14 too, or the project needs to cap/declare the supported Python versions so this deadlock defense isn't claimed on an interpreter where the regression test still fails.

jakepresent (approved):
Looks good to me now. I reran the leaked-worker subprocess regression under 3.11.15, 3.12.3, 3.13.12, and 3.14.3, and it passes on all four. `tests/test_runtime_safety.py` also passes on 3.14.3 (`14 passed`), and `tests/test_runtime_safety.py tests/test_run_metadata.py` passes on 3.11.15 (`32 passed`).

Non-blocking before merge: please refresh the PR body so it matches the current implementation. It still describes the detach as only `_threads_queues`, lists 12 runtime-safety tests, and overstates the watchdog/progress-stall behavior a bit.

PR #23, opened 2026-05-06
refactor(p2m): terminology migration to maximize broad-developer recognition

changliu2:
## Summary

Sweeping terminology migration for the public preview, rebased onto current `main`. The new vocabulary is final — no follow-up rename PR is planned. `prompt` / `scenario` remains the canonical multi-turn discriminator.

Because the diff is large (200+ files), this description categorizes the surface-level changes so a reviewer can verify scope without scrolling the file list.

## Authoring UX (most important user-facing change)

The behavior spec is now a **single self-contained YAML file**. There is no longer a companion `<name>.md` file loaded by the runtime.

```yaml
# Before                       # After
concept:                       behavior:
  name: my_concept               name: my_behavior
# + companion my_concept.md      description: |
#   (markdown loaded at runtime)   Free-form prose that fully replaces
                                   the old companion .md. Authors put
                                   the entire behavior definition here.
```

The 6 stock companion `.md` files were deleted. Authors now write the behavior definition as free-form text directly under `behavior.description`. No magic filename lookup, no second file to keep in sync.

## What the user authors (YAML)

| Before | After |
|---|---|
| `concept.name` + companion `<name>.md` file | `behavior.name` + `behavior.description` (inline; free-form prose) |
| `pipeline.policy` | `pipeline.systematize` |
| `pipeline.seeds` | `pipeline.test_set` |
| `pipeline.rollout` | `pipeline.inference` |
| `factors: [...]` | `test_set.stratify.dimensions: [...]` |
| `kind: prompt \| scenario` (per-record JSONL key) | `type: prompt \| scenario` |
| `examples/concepts/` (authoring directory) | `examples/behavior_specs/` |
| `test_set.tool_source: per_seed` | `test_set.tool_source: per_test_case` (the legacy `per_seed` value is still accepted but emits a `DeprecationWarning`) |

## Pipeline stages (final)

```
systematize -> test_set -> inference -> judge
```

## What p2m generates (artifacts)

| Stage | Artifacts |
|---|---|
| `systematize` | `systematization.json`, `taxonomy.json` |
| `test_set` | `test_set.jsonl`, `stratification.json` |
| `inference` | `inference_set.jsonl` (each row = one test-case execution; shape covers chat conversations AND agent action sequences) |
| `judge` | `scores.jsonl`, `metrics.json` |

## Per-record discriminator (FINAL)

```yaml
type: prompt    # single-turn case
type: scenario  # multi-turn case
```

[...]

PR #63, opened 2026-05-19
Enable Phoenix tracing for Prompt Agent (hosted) targets

tangym:
## Problem

`target.trace: { backend: phoenix }` was silently ignored for Prompt Agent (model-only) targets. The `trace` config was only consumed inside the `if target.is_callable:` branch in `rollout.py`, so `HostedSession` never set up instrumentation.

## Fix

When `target.trace` is set for a hosted target, call `phoenix.otel.register(auto_instrument=True)` before building the session. This detects the installed `openinference-instrumentation-litellm` package and monkey-patches litellm so every `acompletion` call emits spans to Phoenix automatically.

- Idempotent: module-level `_hosted_trace_registered` flag ensures it runs once per process
- Graceful degradation: logs a warning if `arize-phoenix-otel` is not installed
- No new dependencies required (uses existing optional deps)

## Testing

Ran a smoke config with `azure/gpt-5.4-nano` + `trace: { backend: phoenix }`. Confirmed:
- Log: `Enabled Phoenix auto-instrumentation for hosted session tracing`
- 33 traces landed in Phoenix default project
- All rollouts completed successfully

### DIFF
--- a/p2m/stages/inference.py
+++ b/p2m/stages/inference.py
@@ -73,6 +73,38 @@ _INFERENCE_CONFIG_HASH_FILE = ".inference_config_hash"
 
 _VERSIONED_ARTIFACT_RE = re.compile(r"^v\d{4}$")
 
+_hosted_trace_registered = False
+
+
+def _ensure_hosted_trace_instrumentation() -> None:
+    """Register Phoenix tracing so HostedSession LLM calls emit spans.
+
+    HostedSession uses litellm under the hood.  We instrument *only* litellm
+    (via ``openinference-instrumentation-litellm``) to avoid side-effects from
+    auto-instrumenting every installed provider.
+
+    Idempotent: only runs once per process.
+    """
+    global _hosted_trace_registered
+    if _hosted_trace_registered:
+        return
+    try:
+        from phoenix.otel import register
+        from openinference.instrumentation.litellm import LiteLLMInstrumentor
+
+        register()
+        LiteLLMInstrumentor().instrument()
+        _hosted_trace_registered = True
+        log.info("Enabled Phoenix LiteLLM instrumentation for hosted session tracing")
+    except ImportError:
+        log.warning(
+            "target.trace is set but tracing dependencies are not installed. "
+            "Install them for hosted session tracing: "
+            "pip install arize-phoenix-otel openinference-instrumentation-litellm"
+        )
+    except Exception:
+        log.warning("Failed to register Phoenix tracing for hosted sessions", exc_info=True)
+
 
 def _is_versioned_test_set_artifact_path(path: Path) -> bool:
     """Return True if the path lives under a versioned cache directory.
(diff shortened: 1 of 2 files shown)


jakepresent (commented):
This looks directionally good to me. I ran a local compile plus targeted runtime/import tests: `140 passed, 1 skipped`.

I’d wait to approve/merge until #23 lands and this is rebased, since this currently touches `p2m/stages/rollout.py` and #23 moves that surface to `inference.py` / test-case terminology. After the rebase, the only thing I’d want to sanity-check is that the hosted-target trace path still registers Phoenix instrumentation in the renamed inference stage. A tiny unit test around “hosted target + `target.trace` registers once” would be nice, but I don’t think it needs to block if the smoke result holds.

jakepresent:
Found one issue while doing a partial review before the rebase.

The existing targeted tests still pass locally:
- `uv run python -m compileall -q p2m`
- `uv run pytest tests/test_runtime_modes.py tests/test_framework_agnostic.py tests/test_stage_runner_smoke.py -q`
- `137 passed, 1 skipped`

But directly exercising `_ensure_hosted_trace_instrumentation()` crashes in my env. `phoenix.otel.register(auto_instrument=True)` attempts to auto-instrument all installed OpenInference instrumentors, not just LiteLLM. In this checkout it imported the CrewAI/ChromaDB stack and failed with a Pydantic config error before the hosted session was built.

That means `target.trace` on a hosted target can currently turn into a hard runtime crash, even though the intended behavior is graceful degradation. Also, `_hosted_trace_registered = True` is set before registration succeeds, so a failed registration attempt can leave the process in a “registered” state without instrumentation.

I think the fix is either:
1. instrument only LiteLLM explicitly, rather than `auto_instrument=True` across every installed OpenInference integration, or
2. catch broader registration failures, log a warning, and only mark the registration as successful after it actually succeeds.

After that plus the post-#23 rebase from `rollout.py` to `inference.py`, I’m good with the direction.

tangym:
Thanks for the review! Updated with all the fixes:

1. **`auto_instrument=True`** switched to explicit `register()` + `LiteLLMInstrumentor().instrument()` so we control exactly what gets instrumented and avoid the crash you flagged.

2. **Flag timing fixed**: `_hosted_trace_registered` is now set *after* both `register()` and `instrument()` succeed, so a failure mid-setup correctly leaves the flag `False` for retry on next call.

3. **Broader exception handling**: added a general `except Exception` with `exc_info=True` alongside the `ImportError` catch, so unexpected failures from Phoenix/OTel config issues get logged instead of bubbling up.

4. **Rebased onto main**: resolved the renaming from #23.

Also added unit tests (`tests/test_hosted_trace_registration.py`) covering the happy path (register-once idempotency), `ImportError` (missing deps), and `RuntimeError` (config failure).

AaronAspinwall123 (approved):

PR #78, opened 2026-05-22
fix(viewer): support 3-way compare for >2 runs; truncate long callable target labels

changliu2:
## What

Two-part fix for the run-compare page when comparing 3 runs of a target-callable demo (e.g. `private_banking_rm_v3_langchain`, `banking_mcp_langgraph`):
image

1. **3-way compare URLs were not parsed.** The compare page only deserialized the first two `?run=` params. Fixed in `data.ts` to fan out to N runs and in `+page.svelte` to render an N-way layout (was hardcoded to 2-up grid).
2. **Long callable target labels overflowed cards and table headers.** When the target is a Python callable like `examples.banking_mcp_langgraph.agent:chat_shielded`, the prior `runLabel` split on `/` only and returned the entire dotted path. New `runLabel` splits on `:` first (callable targets) then `/` (provider/model paths), so the same target renders as `chat_shielded`. Per-run card title now wraps the label in a `truncate + title` tooltip with `flex-shrink-0` on the color dot.

## Why

Without this, you cannot run the 3-step eval-fix demos (baseline → prompt-hardened → shield) in the viewer compare view, which is the primary visual artifact for //build.

## Validation

- `npm run build` clean
- `pytest viewer/tests` clean
- Manually opened compare URL with 3 runs from `private_banking_rm_v3_langchain` — table headers no longer wrap, color dots stay round, full target string visible on hover.

## Scope

Frontend-only. No schema or data-layer changes.

### DIFF
--- a/viewer/src/lib/server/data.ts
+++ b/viewer/src/lib/server/data.ts
@@ -120,6 +120,7 @@ interface BehaviorComparison {
 	behavior: string;
 	metrics: Record<string, Record<string, CompareMetricSummary>>;
 	deltas: Record<string, number>;
+	deltaByRun: Record<string, Record<string, number>>;
 }
 
 function hasKind(row: Record<string, unknown>, expected: 'prompt' | 'scenario'): boolean {
@@ -862,7 +863,8 @@ function buildBehaviorComparisons(
 				comparisonByBehavior.set(behavior, {
 					behavior,
 					metrics: Object.fromEntries(allMetrics.map((metric) => [metric, {}])),
-					deltas: {}
+					deltas: {},
+					deltaByRun: {}
 				});
 			}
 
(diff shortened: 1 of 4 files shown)


changliu2:
Split into two single-purpose PRs. The label-truncation portion (the part actively blocking PR #88's bank-manager-demo render) lives in the new PR linked above against current main. The 3-way compare URL plumbing portion has been deferred (no current consumer); happy to re-open if anyone needs it.

changliu2:
Reopening per Chang's request. Rebasing onto current main now so this can merge separately ASAP.

changliu2:
Closing as superseded after re-audit against current main.

PR #78's three pieces map to current main as follows:

| PR #78 piece | Current state on main |
|---|---|
| URL plumbing to parse `?runs=v1,v2,v3,v4` | **Already on main** — `viewer/src/routes/suite/[suite_id]/compare/+page.server.ts` parses up to 4 runs since the late-May compare refactor. |
| N-way render loops (was hardcoded 2-up grid) | **Already on main** — `compare/+page.svelte` does `{#each data.runs as run, i}` in 6 places. |
| `runLabel` helper to truncate long callable target labels | **In PR #160** (clean 13-line patch against current main, opened earlier today). |
| Per-run Δ columns in the **By behavior category** table | Genuinely novel; **not on main**. Tracked in #161 as a follow-up enhancement so anyone with bandwidth can pick it up against the new `PrimerDropdown` metric-picker design without inheriting this branch's 158-commit drift. |

PR #160 covers the label-truncation regression that was blocking PR #88's demo render. #161 captures the remaining novel feature so we don't lose the idea. This PR has 158 commits of main drift, which is more work to rebase than to re-author against current main.

PR #87, opened 2026-05-22
Improve error message when viewer encounters outdated artifacts

tangym:
## Summary

Two small fixes for the `p2m` CLI:

### 1. Actionable error message for outdated artifacts

When cached artifacts from a previous run use an older schema (e.g. `kind` instead of `type`), the viewer now shows an actionable error message instead of a generic failure.

**Before**
```
ViewerReadModelBuildError: Missing or invalid kind in /path/to/scores.jsonl
```

**After**
```
scores.jsonl: expected field "type" ("prompt" or "scenario") but the row
contains [auditor_model, concept, judge_error, ...]. The file appears to use
an outdated format that is incompatible with the current version of p2m.
  To fix: delete the run directory
    rm -rf artifacts/results/my-suite/my-run
  and re-run the pipeline so all artifacts are regenerated.
```

- Replace generic errors in `_kind_and_test_case_id` with a `_raise_schema_hint` helper
- Show which field is missing, what was expected, and the actual keys in the row
- Provide a concrete `rm -rf` command pointing at the stale run directory

### 2. Clean shutdown of async tasks (litellm noise fix)

Cancel pending async tasks during event loop teardown in `_bounded_loop_teardown()`. Without this, litellm's internal `LoggingWorker` coroutine gets garbage-collected against a closed loop and produces noisy `Task was destroyed but it is pending!` / `RuntimeError: Event loop is closed` errors on every pipeline run.

- Enumerate and cancel all pending tasks before `loop.close()`, mirroring what `asyncio.run()` does internally
- Validated with 125 live litellm inference calls — no shutdown noise

### Tests
All 58 viewer tests pass.

### DIFF
--- a/p2m/core/runtime_safety.py
+++ b/p2m/core/runtime_safety.py
@@ -448,6 +448,21 @@ def _bounded_loop_teardown(
                     loop.run_until_complete(loop.shutdown_asyncgens())
             except Exception as exc:  # noqa: BLE001
                 err_holder[0] = exc
+            # Cancel all pending tasks (e.g. litellm's LoggingWorker)
+            # before closing the loop. Without this, pending coroutines
+            # get garbage-collected against a closed loop and produce
+            # noisy "Task was destroyed but it is pending!" errors.
+            try:
+                if not loop.is_closed():
+                    pending = asyncio.all_tasks(loop)
+                    for task in pending:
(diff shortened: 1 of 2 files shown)


AaronAspinwall123:
closing for hygiene, reopen if needed

PR #90, opened 2026-05-25
fix: remove stale judge artifacts and warn on incomplete category coverage

tangym:
## Problem

When inference is re-run (via `--force-stage inference` or config change), stale `scores.jsonl` from a previous judge run is left on disk. The viewer then crashes (`ViewerReadModelBuildError`) because scores reference inference rows that no longer exist.

Separately, when `sample_size` is too small relative to the number of behavior categories in the taxonomy, some categories get zero test cases — but the user gets no warning about this.

### 1. Remove stale judge artifacts when inference cache is cleared (`inference.py`)
When inference is forced or its config hash changes, delete `scores.jsonl` and `.judge_config_hash` so a subsequent `p2m results` call never reads outdated scores.

### 2. Gracefully handle stale scores referencing missing inference rows (`viewer_read_model.py`)
Instead of crashing with `ViewerReadModelBuildError`, log a warning and fall back to `transcript_only` mode so the user can still inspect inference outputs.

### 3. Warn when test cases don't cover all behavior categories (`test_set.py`)
After writing `test_set.jsonl`, compare covered categories against the taxonomy. If any are missing, log a `WARNING` with the missing category names and the minimum `sample_size` needed for full coverage.

Example output:
```
WARNING - Test cases cover only 2 of 8 behavior categories. Missing: Cat A, Cat B, ...
         To cover all categories, set sample_size >= 8 per kind (prompt / scenario).
```

## Testing
- All 754 unit tests pass
- Manual verification with azure_doc_qa example (8 categories, sample_size=5 → warning correctly emitted)

### DIFF
--- a/p2m/stages/inference.py
+++ b/p2m/stages/inference.py
@@ -71,6 +71,18 @@ _TESTER_RETRY_GUIDANCE = "Your last reply looked like hidden setup or a scenario
 
 _INFERENCE_CONFIG_HASH_FILE = ".inference_config_hash"
 
+_JUDGE_ARTIFACTS_TO_CLEAN = ("scores.jsonl", ".judge_config_hash")
+
+
+def _remove_stale_judge_artifacts(run_dir: Path) -> None:
+    """Remove judge-stage outputs that depend on the inference data being replaced."""
+    for name in _JUDGE_ARTIFACTS_TO_CLEAN:
+        path = run_dir / name
+        if path.exists():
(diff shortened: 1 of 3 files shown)


jakepresent (approved):
Approving: 52 LOC, both Tier 1 and Tier 4 green, fixes real ViewerReadModelBuildError testers will hit during the bug bash. Merging for tester experience.

Issue #58, opened 2026-05-15
All p2m CLI commands show outdated, inconsistent results vs the viewer; align the CLI output on the viewer's display


changliu2:
## Problem

The `p2m` CLI surface (`p2m results list`, `p2m results show`, `p2m results compare`, `p2m results compare-suites`, `p2m run` end-of-run summary, judge / metrics output) and the SvelteKit viewer in `viewer/` have drifted apart.

The CLI was built incrementally as stages were added; the viewer was built later with a coherent display model for the same artifacts. The result:

- **Different field names** for the same concept (e.g. CLI says one thing, viewer label says another for the same `metrics.json` field).
- **Different aggregation choices** (per-failure-mode rollups, per-dimension rollups, citation rendering) — the viewer's are the considered ones; the CLI's are whatever was easiest at the time.
- **Stale terminology** still present in CLI output (older `risk` / `suite` / `seeds` / `auditor` wording in some places, mixed with the current failure-mode / dataset / test-cases / generation language).
- **Different "default view"** of a run — the viewer leads with verdicts + cited evidence; the CLI leads with raw counts.

A user who runs the CLI then opens the viewer (or vice versa) sees what looks like two different products reporting on the same artifacts. This is a concrete onboarding tax — customers ask "which one is right?"

## Scope (all CLI commands that render results)

- `p2m results list`
- `p2m results show`
- `p2m results compare` (within-suite)
- `p2m results compare-suites` (cross-suite)
- End-of-`p2m run` summary block
- Any judge / metrics rendering surfaced from `p2m/results.py`

## Proposed direction

1. **Adopt the viewer's display model as the source of truth.** The viewer already settled the field names, the aggregation choices, the citation-rendering format, and the verdicts-first narrative. The CLI should render the same model — just in a terminal-clean form.
2. **Extract a shared `display/` module** that both surfaces consume:
   - one place defines field labels (e.g. how `metrics.json` keys are surfaced to a human),
   - one place defines aggregations (per-failure-mode, per-dimension, pass/fail/refusal counts, evidence summaries),
   - one place defines the citation render format.

   CLI gets a Rich-based rendering; viewer gets a Svelte rendering; both pull from the same model.
3. **Pass terminology audit.** Any remaining `risk` / `suite` / `policy` / `seeds` / `auditor` strings in CLI output get migrated to the customer-facing terms (failure-mode / dataset / spec / test cases / generation). Match the viewer.
[...]

PR #92, opened 2026-05-26
Aaspinwall/issue 58 align cli output

AaronAspinwall123:


jakepresent:
Quick review note: the CLI/viewer label alignment itself looks good, and local targeted tests pass:

- python -m pytest tests/test_display_labels.py tests/test_cli.py tests/test_results.py tests/test_runner_stage_filters.py -q

One scope question before I approve: this PR also changes the default inference/judge concurrency from 10 to 25 and adds a new --concurrency CLI flag. That may be useful, but it seems separate from issue #58's CLI-output wording/alignment. Was that intentional for this PR, or should it be split into a small follow-up so this one stays display-only?

AaronAspinwall123:
> Quick review note: the CLI/viewer label alignment itself looks good, and local targeted tests pass:\n\n- python -m pytest tests/test_display_labels.py tests/test_cli.py tests/test_results.py tests/test_runner_stage_filters.py -q\n\nOne scope question before I approve: this PR also changes the default inference/judge concurrency from 10 to 25 and adds a new --concurrency CLI flag. That may be useful, but it seems separate from issue #58's CLI-output wording/alignment. Was that intentional for this PR, or should it be split into a small follow-up so this one stays display-only?

Yes, it was deliberate to add the concurrency bump. I was working on this one and remembered I needed to add the concurrency flag and bump the default, so I just added it here.

jakepresent:
I think the `--concurrency` flag is probably fine, but the default bump from 10 to 25 is a real behavior change, not just PR hygiene. It changes fan-out for any config that doesn’t set concurrency and could affect provider quotas/flakiness. I’d prefer keeping #92 focused on CLI label alignment, and moving the default bump to a small follow-up with its own rationale/evidence.

AaronAspinwall123:
> I think the `--concurrency` flag is probably fine, but the default bump from 10 to 25 is a real behavior change, not just PR hygiene. It changes fan-out for any config that doesn’t set concurrency and could affect provider quotas/flakiness. I’d prefer keeping #92 focused on CLI label alignment, and moving the default bump to a small follow-up with its own rationale/evidence.

Okay will leave default 10 for now

tangym (approved):

tangym (approved):

Issue #97, opened 2026-05-26
Many "Grouped by behavior category" views and stacked descriptive text in eval run page


minthigpen:
Pulled latest from main. Ran example pipes/health_assistant.yaml and when I click into the eval run, there is a weird stacking of the descriptive text, and "Grouped by behavior category" is repeated multiple times with many different filters.

<img width="2484" height="1272" alt="Image" src="https://github.com/user-attachments/assets/71d0d826-f927-4692-a73a-81dd9797d027" />

All the different "Grouped by behavior category" views. Expected is a single "Grouped by behavior category" view with all different variations of policy status. It seems to be pre-filtering them.

<img width="1796" height="841" alt="Image" src="https://github.com/user-attachments/assets/a0590844-adba-4764-9626-e15b4b3be7b5" />

<img width="1761" height="796" alt="Image" src="https://github.com/user-attachments/assets/ded2391a-6bae-4a49-bf49-e1e93e22666e" />

<img width="1839" height="995" alt="Image" src="https://github.com/user-attachments/assets/8855fa65-d67e-440e-a8da-5cfe83147b85" />

<img width="1792" height="982" alt="Image" src="https://github.com/user-attachments/assets/745539ee-244d-4a40-ad1a-72466788908d" />

AaronAspinwall123:
<img width="2051" height="846" alt="Image" src="https://github.com/user-attachments/assets/d448ef48-d02c-4abc-bc62-6dd1c7ed4cc3" />

Fixed with this PR:

https://github.com/microsoft/ASSERT/pull/105

Replaced the series of tabs with a drop down menu. Strengthened protection for long description text to not overflow.

PR #105, opened 2026-05-26
fix(viewer): repair run page grouping tabs (ASSERT #97)

AaronAspinwall123:
The run page rendered duplicate "Grouped by behavior category" tabs because the segmented control hardcoded the label instead of using each axis's own label, and buildFactorAxes() generated an O(N^2) explosion of single + pairwise axes.

Changes:
- grouping.ts: rename observed_node axis label to "Behavior category" and drop the pairwise axis loop so we only emit one axis per dimension (linear growth).
- run page: replace the SegmentedControl with a "Group by" PrimerDropdown placed next to the existing "Filter by" dropdown in both prompt and audit sections, so the header scales to any number of dimensions.
- run page: restructure the section header so the H2 + description stack above the controls instead of sharing a row, and add truncate / line-clamp-2 so long titles and descriptions can never collapse the layout.

image

### DIFF
--- a/viewer/src/lib/grouping.ts
+++ b/viewer/src/lib/grouping.ts
@@ -88,7 +88,7 @@ function observedNodeSort(a: GroupEntry<ScoredRecord>, b: GroupEntry<ScoredRecor
 // ---------------------------------------------------------------------------
 
 const GROUP_AXES: GroupAxis<ScoredRecord>[] = [
-	{ key: 'observed_node', label: 'Observed behavior (judge)', accessor: observedNodeAccessor, sortGroups: observedNodeSort },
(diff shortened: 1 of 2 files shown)


jakepresent (approved):
Looks good. This addresses the run-page grouping issue by using axis labels instead of a repeated hardcoded label, removing the pairwise-axis explosion, and moving grouping into dropdown controls that scale better.\n\nValidation:\n- npm --prefix viewer run build\n- python -m pytest tests/test_viewer_run_page_server.py tests/test_viewer_read_model_turns.py tests/test_viewer_server_artifacts.py -q\n\nNote: npm --prefix viewer run check still reports pre-existing viewer type/a11y issues in PrimerPagination, +page.svelte, new/+page.svelte, and PrimerDropdown; not introduced by this PR.

Issue #100, opened 2026-05-26
Create run: searching for pre-existing behaviors shows behavior categories


minthigpen:
When user clicks to search for existing behaviors, these show the generated behavior categories -- not the actual behavior itself which is harmful medical advice. 

<img width="1591" height="890" alt="Image" src="https://github.com/user-attachments/assets/41a23ba1-01e2-4dce-8fe1-d91bea41b522" />

AaronAspinwall123:
https://github.com/microsoft/ASSERT/pull/106/changes

PR to fix it: 

<img width="1401" height="897" alt="Image" src="https://github.com/user-attachments/assets/747a4eba-5fee-4d5f-87c1-c17070b7fc94" />

PR #106, opened 2026-05-26
fix(viewer): return behavior specs, not categories, from /api/behavio…

AaronAspinwall123:
…rs (ASSERT #100)

The Create-run wizard's "Search behaviors" dropdown was listing the generated behavior categories from each suite's taxonomy (taxonomy.behavior_categories), not the actual behavior specification (taxonomy.behavior). For a suite about harmful medical advice, the user saw sub-categories like "Recommending harmful drugs" instead of the underlying behavior itself.

Switch the endpoint to emit one entry per suite using taxonomy.behavior (name + definition), still dedup-ed by behavior name across suites.

image

### DIFF
--- a/viewer/src/routes/api/behaviors/+server.ts
+++ b/viewer/src/routes/api/behaviors/+server.ts
@@ -8,16 +8,14 @@ export const GET: RequestHandler = async () => {
 
 	for (const suite of suites) {
 		const taxonomy = loadPolicy(suite.suite_id);
-		if (!taxonomy?.behavior_categories) continue;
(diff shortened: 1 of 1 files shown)


jakepresent (approved):
Looks good. This fixes the create-run search at the right layer by returning one behavior spec per suite instead of generated behavior categories, while preserving the existing de-dupe behavior.\n\nValidation:\n- npm --prefix viewer run build\n- python -m pytest tests/test_viewer_policy_api.py tests/test_viewer_server_artifacts.py -q

Issue #104, opened 2026-05-26
Evaluating model target should not require tools


minthigpen:
In the create flow, first step: Input specification, if user selects model, they should not see required section for tools. Remove. 

<img width="1473" height="1301" alt="Image" src="https://github.com/user-attachments/assets/357262df-34ef-4e81-82e9-b4f33b0e2353" />

Also, system prompt is required for a model, remove "(optional)" if model is selected. If "Prompt Agent" is selected as an evaluation target, then it should be "optional" but right now the text field for System prompt disappears completely.

<img width="1413" height="725" alt="Image" src="https://github.com/user-attachments/assets/b54b635e-e39e-404d-83b5-3e4e3e76154f" />

<img width="1179" height="702" alt="Image" src="https://github.com/user-attachments/assets/dfbbd345-dcbf-4b0a-9896-4da635d0db20" />

AaronAspinwall123:
Fixed here: 

https://github.com/microsoft/ASSERT/pull/107

<img width="1539" height="1243" alt="Image" src="https://github.com/user-attachments/assets/65ad2105-ecc7-495e-a2d5-2e301d1659d1" />

<img width="1506" height="1150" alt="Image" src="https://github.com/user-attachments/assets/6566db11-ce98-4ee3-8472-482ab09eca5d" />

PR #107, opened 2026-05-26
fix(viewer/new): align Step 1 target fields with target type (ASSERT …

AaronAspinwall123:
…#104)

Create-run Step 1 had two related issues:
- Tools section was always shown, including a required field, even when the user picked "Model" as the evaluation target.
- The System prompt field was conditionally rendered: shown (as optional) for Model, hidden entirely for Prompt Agent. The opposite is correct -- the prompt is required for Model and optional for Prompt Agent.

Changes:
- Always render the System prompt input. Show a required asterisk and validation message when the target is Model; show "(optional)" with a Prompt-Agent-appropriate hint otherwise.
- Hide the entire Tools section (and skip its validation) when the target is Model. It now only renders for Prompt Agent.
- Include the system prompt in the submitted payload and Step 3 summary regardless of target, so an operator-provided prompt for a Prompt Agent target is preserved.

image
image

### DIFF
--- a/viewer/src/routes/new/+page.svelte
+++ b/viewer/src/routes/new/+page.svelte
@@ -299,12 +299,18 @@
 			: newBehavior.name.trim().length > 0 && newBehavior.definition.trim().length > 0
 	);
 	let step1ContextValid = $derived(applicationContext.trim().length > 0);
-	let step1ToolsValid = $derived(
-		toolsMode === 'simulated'
+	let step1ToolsValid = $derived.by(() => {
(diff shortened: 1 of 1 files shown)


jakepresent (approved):
Looks good. The field visibility/validation now matches the selected target: model targets require a system prompt and skip tools, while Prompt Agent targets keep tools and preserve an optional system prompt.\n\nValidation:\n- npm --prefix viewer run build\n- python -m pytest tests/test_viewer_runner.py tests/test_viewer_server_artifacts.py -q

PR #122, opened 2026-05-28
Aaspinwall/issue 98 overrefusal UI

AaronAspinwall123:


AaronAspinwall123:
image
New UI look

minthigpen:
If the target refused the input or refused before inference then these should not proceed with being scored for overrefusal and policy violation. Could we have some logic to ensure that these not be scored in the pipeline?

AaronAspinwall123:
> If the target refused the input or refused before inference then these should not proceed with being scored for overrefusal and policy violation. Could we have some logic to ensure that these not be scored in the pipeline?

Thanks @minthigpen 

That makes sense. I update the logic and these dont proceed anymore. I am still showing them in the UI, but I can also fully remove them if thats better? 

Current look: 
image

minthigpen:
This is much better, no I think we should include them still so that users can see which rows failed. Thanks! #shipit

tangym (approved):
1. Duplicated UI helper functions

`stopReasonLabel()`, `stopReasonTitle()`, and `stopReasonChipClass()` are copy-pasted between `ResultDrawer.svelte` and `+page.svelte`. These should be extracted to a shared module (e.g., `result-view.ts` where `scenarioStopReasonDisplay` already lives) to avoid future divergence.

2. Potential double-counting in refused/errored banner

In `+page.svelte`, `auditRefusedCount` and `auditErroredCount` sum across **both** `data.auditScores` and `data.inferencePreviewRows`:

```ts
let auditRefusedCount = $derived(
    data.auditScores.filter(
        (s) => s.metadata?.stop_reason_display?.tone === 'refusal'
    ).length +
        (data.inferencePreviewRows ?? []).filter(
            (r) => r.stop_reason_display?.tone === 'refusal'
        ).length
);
```

If a scenario appears in both collections (scored run that also has preview data), this would double-count. Worth verifying these are mutually exclusive or deduplicating by `test_case_id`.

3. `runtime_close_error` asymmetry between backend and viewer

The viewer maps `runtime_close_error` to `tone: 'error'` but the backend judge does **not** include it in `_UNSCORABLE_STOP_REASONS` — these rows still go to the judge LLM. The test explicitly lists it as "scorable", so this seems intentional (the conversation did complete, just the cleanup failed), but the error tone in the UI might confuse users into thinking it was skipped. Worth a comment in `scenarioStopReasonDisplay` explaining why this is scored despite the error tone.

4. `normalizeAuditScore` now allocates on every call

This was previously a no-op passthrough (`return score`). Now it spreads a new object on every audit score:

```ts
function normalizeAuditScore(score: AuditScore): AuditScore {
    const stopReason = typeof score.metadata?.stop_reason === 'string' ? score.metadata.stop_reason : '';
    return {
        ...score,
        metadata: {
            ...score.metadata,
            stop_reason_display:
                score.metadata?.stop_reason_display ?? scenarioStopReasonDisplay(stopReason)
        }
    };
}
```

For large result sets this adds allocation pressure. Consider short-circuiting when `stop_reason_display` is already present:

```ts
if (score.metadata?.stop_reason_display != null) return score;
```

tangym (approved):

PR #134, opened 2026-05-29
Refine init system prompt: pacing, presets, and max-turns

tangym:
## Summary

Addresses post-merge feedback from @minthigpen on PR #110 and fixes several issues discovered during testing with `gpt-5.4-mini`.

## Changes

1. **Refine init conversation flow** (`f9f17719`) — Apply minthigpen's feedback: simplify Judge Configuration section, streamline System Context guidance, tighten Behavior Definition instructions, and clean up Test Set Generation wording.

2. **Add section-coverage pacing guardrails** (`304fe642`) — Prevent mini models from rushing to `propose` by adding explicit pacing rules: touch all 5 sections before proposing, acknowledge info from rich answers but continue to the next uncovered section.

3. **Move checklist into propose action** (`33d75522`) — Gate `propose` behind a 5-point prerequisite checklist. The LLM must verify all sections are covered before emitting a proposal.

4. **Don't ask for system prompt when target is callable** (`2c80fa2c`) — Callable targets (agent frameworks) own their own prompts. The init flow now skips the system prompt question for callables.

5. **Make judge presets explicitly optional** (`ae054113`) — Rewrite Section 5 to separate presets from custom dimensions, and instruct the LLM to omit `preset:` from YAML when the user declines. Fixes mini models hallucinating presets when user says "none".

6. **Increase default --max-turns from 20 to 30** (`e55ff1f2`) — With 5 question sections + propose/refine cycles + validation retries, 20 turns was too tight.

## Testing

Tested with `assert-eval init --dry-run --model azure/gpt-5.4-mini` and `azure/gpt-5.4`.

### DIFF
--- a/assert_eval/init/_command.py
+++ b/assert_eval/init/_command.py
@@ -73,7 +73,7 @@ log = logging.getLogger(__name__)
 @click.option(
     "--max-turns",
     type=int,
-    default=20,
+    default=30,
     show_default=True,
     help="Maximum conversation turns.",
 )
--- a/prompts/init_system.md
+++ b/prompts/init_system.md
@@ -56,9 +56,12 @@ Help select one:
 Help users focus on dimensions tied to **real failure modes**
 
 ### 5. Judge Configuration
-First inform users that built-in judge dimensions (policy violations and overrefusal) will be used. But ask users if they'd like to optionally configure custom judge dimensions. If so, give guidance that users should
-  - define clear, observable, actionable criteria
-  - ensure dimensions help debugging and interpretation for tradeoffs
+First inform users that two built-in judge dimensions (policy violations and overrefusal) are always included automatically — no configuration needed. Then ask two separate questions:
+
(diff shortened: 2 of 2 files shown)


AaronAspinwall123 (approved):
Approving with some non-blocking copilot comments

 1. done can bypass the new pacing guardrail.
 The new checklist says not to emit "propose" until all 5 sections are covered, but the runtime accepts "done" with YAML the same way. A model could still rush directly to "done" in interactive mode and bypass the guardrail. The prompt should gate both "propose" and "done", or say "done" is only for non-interactive/final accepted configs.
 2. New “decline custom dimensions” rule conflicts with existing guidance.
 The added section says to omit dimensions: if the user declines custom dimensions, but later guidance still says to “Default to 2-4 custom judge dimensions.” That can cause the exact hallucinated-dimensions behavior this PR is trying to prevent. The later line should be narrowed to cases where the user wants custom dimensions or non-interactive inference is needed.

PR #137, opened 2026-05-29
Auto-detect OTel collector and improve logging

tangym:
## Summary

Suppress noisy gRPC exporter errors when no OTel collector is running, clean up logging across the pipeline, and add a viewer deep-link to end-of-run output.

## Problem

When running without an OTel collector, the gRPC exporter spams errors like:
```
Failed to export batch. Reason: …Connection refused
```
These are confusing to users who did not opt into tracing.

Additionally, log output from ASSERT's own dependencies (alembic, phoenix, opentelemetry SDK) was noisy, and the end-of-run summary only showed a CLI command — not the web viewer.

## Changes

1. **Auto-detect OTel collector** (`otel.py`): Probe `localhost:4317` before configuring the gRPC exporter. If unreachable, skip export silently instead of retrying and flooding stderr.

2. **Promote collector detection to INFO** (`otel.py`): Log whether a collector was detected or skipped at INFO level so users know what happened.

3. **Suppress noisy dependency loggers** (`logging_config.py`): Add alembic, phoenix, and opentelemetry SDK loggers to the suppression list. Only ASSERT's own deps are suppressed — user agent deps (crewai, langchain, etc.) are left alone.

4. **Add `[stage]` prefix to stage log messages** (`inference.py`): Pipeline stages use a `[stage_name]` prefix for clarity. Removed `[otel]` prefixes from non-stage logs to keep prefixing consistent.

5. **Suppress opentelemetry SDK logger** (`logging_config.py`): The OTel SDK emits verbose warnings even in normal operation.

6. **Add viewer deep-link to end-of-run output** (`runner.py`): Show a direct URL to the viewer for the completed run so users can one-click to visualized results.

## Testing

- 821 tests passed, 14 skipped, 1 pre-existing failure (unrelated PermissionError)
- Verified with and without a local OTel collector running
- Confirmed end-of-run output shows correct deep-link URL

### DIFF
--- a/assert_eval/core/otel.py
+++ b/assert_eval/core/otel.py
@@ -20,9 +20,14 @@ from __future__ import annotations
 
 import asyncio
 import json
+import logging
+import os
+import socket
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Protocol, runtime_checkable
+
+log = logging.getLogger(__name__)
 
 
 # OpenInference semantic conventions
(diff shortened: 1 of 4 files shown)


AaronAspinwall123 (approved):
Seems good to me. I run a copilot review and got some non-blocking comments

 1. otel.py: include OTEL_EXPORTER_OTLP_TRACES_ENDPOINT. The explicit-env check handles OTEL_EXPORTER_OTLP_ENDPOINT but misses the trace-specific standard var, so a user-configured trace exporter can be replaced silently.
 2. otel.py: validate PHOENIX_GRPC_PORT. isnumeric() + int() allows out-of-range ports like 99999; socket.create_connection() can raise outside the current OSError handling. Parse with try/except and enforce 1 <= port <= 65535.
 3. otel.py: probe/logging runs for standard SDK providers too. _should_preserve_default_exporter() runs before the TypeError fallback, so standard OTel providers can get misleading “trace export disabled/exported” logs even though the kwarg is ignored and existing processors remain unchanged. Minor UX issue.
 4. logging_config.py: setLevel(WARNING) won’t suppress the warning named in the comment. WARNING-level logs still pass. If the goal is the specific “Overriding of current TracerProvider” warning, use a targeted filter; setting the whole opentelemetry tree to ERROR would risk hiding useful warnings.
 5. runner.py: viewer hint only works in a prepared source checkout. cd viewer && npm run dev assumes the viewer/ directory exists, dependencies are installed, and the viewer’s default ARTIFACTS_ROOT matches the run. Pip-installed users or custom results_dir/artifacts_root users may get a broken hint.
 6. Tests: worth adding small unit tests for env-var precedence, collector probe behavior, invalid PHOENIX_GRPC_PORT, and replace_default_processor behavior.

PR #142, opened 2026-05-29
fix: restore legacy CLI aliases

jakepresent:
## Summary
- restores the legacy `assert` and `p2m` console scripts as aliases to `assert_eval.cli:cli`
- adds a small metadata test so future rename cleanup does not accidentally drop the compatibility aliases again

## Validation
- `python3 -m pytest tests/test_package_metadata.py tests/test_import_smoke.py tests/test_cli.py -q`
- parsed `pyproject.toml` directly and verified all three console scripts point to `assert_eval.cli:cli`
- `git diff --check`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,6 +92,8 @@ index-url = "https://pypi.org/simple"
 
 [project.scripts]
 assert-eval = "assert_eval.cli:cli"
+assert = "assert_eval.cli:cli"
+p2m = "assert_eval.cli:cli"
 
 [build-system]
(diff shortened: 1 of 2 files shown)


jakepresent:
Withdrawing this. After re-checking current main, the rename is intentionally enforcing assert-eval as the only supported CLI entry point, and tests/test_no_p2m_references.py is designed to prevent reintroducing p2m. Adding aliases would fight that direction rather than preserve an agreed compatibility contract.

PR #146, opened 2026-05-30
Auto-fallback from Azure Responses API to Chat Completions for regions without Responses support

### LINKED ISSUE #129: bug: gpt-5 deployments outside Responses-API-enabled regions (e.g. West Europe) fail with 'Responses API is not enabled' — should auto-fallback to Chat Completions
## Repro

1. Configure a `.env` with an Azure OpenAI deployment in a non-Responses-API region (e.g. West Europe):
   ```
   AZURE_API_BASE=https://<your-westeurope-deployment>.openai.azure.com/
   AZURE_API_KEY=<key>
   ```
2. Run any ASSERT config using `azure/gpt-5.4` as the systematize / judge / inference model
3. Pipeline fails at the first stage that hits Responses API:

```
INFO     [systematize] [1/2] Researching behavior taxonomy...
WARNING  azure/gpt-5.4 (attempt 1/6), retrying in 1.1s: Provider
         error: litellm.APIError: AzureException APIError -
         {""error"":{""code"":""NotFound"",""message"":""Azure OpenAI Responses API is not enabled in this region. Please check
         https://aka.ms/aoai/responsesapi/availability""}}
```

All 6 retries hit the same region → run aborts.

## Why this matters

Azure OpenAI Responses API is currently enabled only in **East US, East US 2, Sweden Central, and South India** (per https://aka.ms/aoai/responsesapi/availability). Any developer with a GPT-5 deployment in any other region — including **West Europe**, where many EU-hosted enterprise deployments live — hits this on first run.

Today this looks like a config bug to the developer (no obvious next step from the error). It is actually a regional capability gap with a known working fallback: Azure OpenAI Chat Completions supports the same GPT-5 model family in every region the model is deployed.

LiteLLM defaults to Responses API for `azure/gpt-5.*` because that is the OpenAI-side default for the GPT-5 family. ASSERT inherits that default with no fallback path.

## Suggested fix

On the first failed call with this specific error signature, auto-fallback to Chat Completions for the remainder of the run, with a single WARN-level log:

```
WARN  azure/gpt-5.4: Responses API not enabled in <region>; falling back to Chat Completions for this run.
      To suppress, set ASSERT_PREFER_CHAT_COMPLETIONS=1 or use a Responses-API-enabled region.
      See: https://aka.ms/aoai/responsesapi/availability
```

Same shape as the existing token / rate-limit warnings: silent, opinionated, surfaces the gap once without blocking the pipeline.

## Acceptance criteria

- [ ] `azure/gpt-5.*` runs in non-Responses-API regions complete end-to-end without manual config
- [ ] WARN is emitted exactly once per run, not per call
- [ ] An env-var override exists for developers who want to force one transport
- [ ] Full error remains visible in DEBUG-level logs (don't swallow)

## Workaround until fix lands

Set `LITELLM_USE_RESPONSES_API=false` in `.env` or pin the model to a Responses-API-enabled region. Worth documenting in `docs/troubleshooting.md` as an interim.

## Source

Surfaced by Hua Liang in the 2026-05-28 internal UX testing bug bash.

tangym:
Closes #129.

## Problem

When running `assert-eval` against an Azure OpenAI endpoint in a region that doesn't yet support the Responses API (e.g. West Europe, or any deployment on an older API version), the framework crashed mid-pipeline with errors like:

```
openai.BadRequestError: Responses API is not enabled for this deployment
openai.BadRequestError: API version 2024-... is not supported by Responses API
```

This blocked every stage that goes through `model_client.generate*` — including `systematize`, `stratification`, and the judge.

## Solution

Layered fallback that detects Responses-API rejections, drops to Chat Completions for the rest of the run, and degrades `web_search` gracefully so pipeline stages keep working instead of crashing.

### Phase 1 — Core fallback (Responses API → Chat Completions)

- Detect Responses-API rejection from the provider error message (covers both "not enabled for this deployment" and "API version … not supported").
- Raise a typed `_ResponsesApiNotAvailableError` (subclass of `LLMProviderError`) so retry logic and tests can distinguish it from generic provider errors.
- Activate a process-wide "force chat completions" flag on the first such error so subsequent calls skip the Responses round-trip entirely.
- Monkey-patch the LiteLLM bridge-check (`responses_api_bridge_check`) so LiteLLM stops trying to route `acompletion` calls through the Responses API once the fallback is active.
- Honor `ASSERT_PREFER_CHAT_COMPLETIONS=1` at import time so users who already know their region lacks Responses can skip the initial probe + fallback round-trip.

### Phase 2 — Robustness

- **Per-task retry budget**: each `generate*` call gets its own `chat_fallback_attempts` budget so a long-running pipeline doesn't compound retries across tasks after the fallback is activated.
- **Classifier de-gating**: the "Responses API not enabled" / "API version not supported" markers are recognized regardless of whether the fallback flag is already set — fixes a case where a second concurrent task could miss the marker.
- **AZURE_API_BASE normalization**: defensively strip a trailing `/openai` path suffix from `AZURE_API_BASE` (LiteLLM appends it itself; a doubled path was producing 404s on some endpoints).

### Phase 3 — `web_search` graceful degradation

When fallback is active, the `web_search` tool is dropped (Chat Completions doesn't support `web_search_preview`) and the call proceeds without web grounding:

- **Proactive** drop at the top of `generate()` / `generate_structured()` when `_force_chat_completions` is already set and `options.web_search=True`.
- **Reactive** drop on `_ResponsesApiNotAvailableError`, with a single retry that recurses with `web_search=False`.
- One-time WARN per run (gated by `_web_search_drop_warned` module flag) so logs aren't spammed:
  > `dropping web_search and routing via Chat Completions (<reason>). Web grounding is disabled for the remainder of this run.`
- Affects only the two `web_search=True` sites in the framework: `stages/systematize.py` (taxonomy research) and `stages/stratification.py` (policy stratification). Both still produce useful output without web grounding.

## Testing

- 38/38 model_client tests pass.
- New `WebSearchFallbackDegradationTest` (4 cases) covers proactive drop, reactive drop, one-time WARN, and the no-op case when `web_search=False`.
- Existing tests for Responses-API classifier ordering and the fallback gate still pass.
- Manually verified end-to-end against an Azure OpenAI endpoint in West Europe (no Responses API): `systematize` and `stratification` complete, WARN is logged once, no crash.

## Diff scope

3 files: `assert_eval/core/model_client.py`, `assert_eval/init/_llm.py`, `tests/test_model_client.py` (+754 / -5).

## Follow-ups (filed separately)

- Make Responses API fallback / `web_search` degradation observable (`metrics.json` + viewer banner) and controllable (`--strict-web-search` / `ASSERT_STRICT_WEB_SEARCH=1`) — #149


AaronAspinwall123 (commented):
1. LiteLLM monkey-patch is signature-coupled and fragile. _guarded_bridge_check hardcodes (model, custom_llm_provider, web_search_options=None, tools=None, reasoning_effort=None). I confirmed it matches the live litellm 1.79.1+ signature today — but if LiteLLM adds a kwarg in a minor release (they do this often), the wrapped function silently drops it and Responses API routing breaks for everyone who wants it. Easy fix: accept **kwargs and forward them.
 2. Import-time mutation of os.environ in _normalize_azure_api_base() is surprising for library users (anyone importing assert_eval.core.model_client gets their env rewritten). It's defensible for a CLI but should at least log at INFO before mutating (it does ✓). Bigger concern: _activate_chat_completions_fallback at import time installs the monkey-patch on litellm.main — importing this module globally mutates LiteLLM. Worth a one-line note in the module docstring.
 3. Test state leakage. WebSearchFallbackDegradationTest's reactive test triggers a real call to _install_responses_api_guard, which monkey-patches litellm.main.responses_api_bridge_check and never unpatches. Subsequent tests run against a patched LiteLLM. Today this is benign because _force_chat_completions is False after tearDown, but it's a smell. Add a _responses_api_guard_installed reset (and original-function restore) in tearDown.
 4. PR is "dirty" against main — though my local merge succeeded with no conflicts. The "dirty" state is likely stale; a rebase will also clear the 28 stale test_viewer_server_artifacts.py failures (file was deleted on main). Recommend squash-merge given 18 commits with reverts/iterative discovery.
 5. init/_llm.py duplicates the one-shot fallback pattern (lines 46–66) without the per-task budget. Fine for the single-shot init agent, but the comment should call out that the duplication is intentional.

Few comments from copilot.

AaronAspinwall123 (approved):

PR #159, opened 2026-05-30
examples(bank-manager-agent-shield): add demo with v3 guardrails

Spiderpc:
Self-contained AgentShield demo under `examples/bank_manager_agent_shield/` (no changes outside the folder).

## What's included

- LangGraph ReAct bank-manager agent with mock banking MCP server (`agent.py`, `mcp_server.py`)
- Three guardrail variants: unguarded baseline, v2 (Stage 2 state gates), v3 (sensitivity-scoped Stage 3 patterns + Stage 4 prompt-injection scrubber)
- 100-case eval suite (50 prompt + 50 scenario) with two custom judge dims (`safety_violation` + `unjustified_refusal`) that bypass the built-in PV/OR derivation
- Frozen n=100 results for the unguarded baseline and v3 variant under `results/` so the viewer renders them out of the box

## Results (n=100)

| run | safety_violation | unjustified_refusal |
|---|---|---|
| unguarded | 39% | 2% |
| **v3** | **1%** | **2%** |

v3 cuts safety failures by 97% (39 \u2192 1) with no helpfulness cost vs the unguarded baseline.

PR #164, opened 2026-05-31
examples(bank-manager): add new-ACS-guarded variant alongside baseline

Spiderpc:
Stacks on top of PR #159. Adds a fourth callable target `chat_guarded_acs` that gates the LangGraph bank-manager agent with the new Agent Control Specification (ACS) runtime, alongside the existing unguarded / v2 / v3 variants.

## What's new

- `acs/manifest.yaml` — binds Rego policy to `input`, `pre_tool_call`, `post_tool_call` intervention points.
- `acs/policy/bank_manager.rego` — stateless policy mirroring v3 semantics (SSN input regex, sensitivity-scoped ACC-1002/1003 gates, approval/admin-mode gates, generic prompt-injection scrubber on tool results).
- `chat_guarded_acs` callable that wraps every MCP tool with a per-call ACS pre/post evaluation. The host tracks per-turn state (`transfer_approved`, `admin_mode_active`, `account_sensitivity`) and threads it into each snapshot, since the new ACS runtime is stateless.
- `eval_guarded_acs.yaml` — comparison config reusing the baseline suite-root test_set + the same safety_violation/unjustified_refusal judge dimensions.

## How this is additive

Nothing existing is changed. The ACS SDK import is wrapped in try/except so the other three variants still work when `agent_control_specification` is not installed. `chat_guarded_acs` raises a clear install message at call time if the SDK is missing.

## Runtime requirements (new)

- `agent_control_specification` Python SDK (built from local AgentControlSpecification repo via maturin; not yet on PyPI).
- `opa` binary on PATH.

See the README's 'Optional: ACS variant' section for setup.

## Validation

- Module imports cleanly with no SDK installed (verified).
- `opa check` passes on the Rego.
- 7 `opa eval` cases cover each gate: sensitive read deny, standard read allow, create_transfer with/without approval, SSN input deny, clean input allow, injected memo deny, enable_admin_mode deny.

## What's not in this PR

Live n=100 results — local install of the SDK requires the MSVC linker, which is being sorted out separately. The viewer-renderable results will be added in a follow-up commit once the SDK builds.

### DIFF
--- a/examples/bank_manager_agent_shield/README.md
+++ b/examples/bank_manager_agent_shield/README.md
@@ -6,22 +6,28 @@ and scores three variants against the same 100-case eval suite.
 
 ## What's here
 
-- `agent.py` — three ASSERT callable targets (`chat_unguarded`,
-  `chat_guarded_v2`, `chat_guarded_v3`) over the same LangGraph ReAct
-  agent.
+- `agent.py` — four ASSERT callable targets (`chat_unguarded`,
+  `chat_guarded_v2`, `chat_guarded_v3`, `chat_guarded_acs`) over the
+  same LangGraph ReAct agent.
 - `mcp_server.py` — mock banking MCP server (`read_account`,
   `read_transaction_history`, `prepare_transfer`,
   `request_customer_approval`, `create_transfer`, `freeze_account`,
   `enable_admin_mode`).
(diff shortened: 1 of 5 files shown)


Spiderpc:
Closing — not ready to PR yet.

PR #180, opened 2026-06-01
viewer/compare: pin baseline left + show only populated dim columns

changliu2:
# viewer/compare: pin baseline left + show only populated dim columns

## TL;DR

Three independent UX/correctness fixes on the compare and suite-landing views, in three commits:

1. **`viewer/suite`** — render columns from the eval's actual `dimensionDefs` instead of three hardcoded headers (`Policy violation`, `Overrefusal`, `Harm actionability`). For evals with custom dimension names (e.g. `safety_violation`, `unjustified_refusal`), the previous UI silently relabeled the customer's metric and showed a dead `—` column for the unsupported `harm_actionability`. Headers now come from `Object.keys(data.dimensionDefs)`, per-row values from `metrics.dimensions[name].rate`, and any column where every visible row is null is hidden. No cap — the table scrolls horizontally like the by-behavior heatmap.

2. **`viewer/compare`** — pin the baseline run to the leftmost column across summary cards, the by-behavior heatmap header, and per-prompt sample columns. Tagging a card `Baseline` while leaving it on the right inverted the diff-tool convention (baseline left, change right) every reader expects; `-18%` deltas read as "left is better than right" on first glance when they actually mean "left is better than baseline on the right." Colors are now keyed by `run_id` via a `runColor` map so a run keeps its color regardless of column position. `?runs=` URL order is preserved as the canonical user-selection order.

3. **`viewer/compare`** — sort the per-prompt sample panels with a three-key comparator: (baseline-flagged DESC, has-disagreement DESC, |Δ| DESC), tiebreak on `prompt`. Lead with the regression-fix story (baseline-fail / variant-clear at top) without hiding regressions (baseline-clear / variant-fail next). The previous sort was disagreement-first then `scoreSortValue` on whichever sample happened to be first non-null, which made screenshot order subtly non-deterministic when the baseline was missing a prompt.

## Decisions worth flagging

- **No column cap on the suite-landing dim table.** The spec floated capping at 5; dropped because the by-behavior heatmap already overflows horizontally and a cap forces an overflow UI we don't need. If a customer has >7 dims we'd rather they scroll than introduce a kebab-menu column picker now.
- **Sort key 1 is `baseline-flagged` = `getRecordFlag(sample, metric) === true`.** I treated judge-failure rows as not-flagged so they fall to the bottom (rather than being treated as agreement-on-clear). Matches the per-spec acceptance criterion.
- **Aggregate helpers `aggregateRunViolationRate` / `aggregateRunOverrefusalRate` are left in place** even though the table no longer calls them — they're tiny and may have downstream callers. Will sweep in a follow-up if confirmed dead.

## How to verify

Existing bank-manager ACS demo artifacts in `artifacts/results/bank-manager-agent-shield/` exercise all three changes:

- Open the suite landing page → the table should show columns for `safety_violation` and `unjustified_refusal` (the eval's actual dims) rather than the old hardcoded three. No empty `Harm actionability` column.
- Click Compare → pick a baseline → it should render leftmost across the cards, the heatmap header, and inside any expanded behavior row's sample panel.
- Expand a behavior with mixed outcomes → baseline-flagged-variant-clear prompts (the "wins") should be at the top; regressions next; agreement at the bottom.

## CI / typecheck

`npm run check` in `viewer/` reports the same 3 pre-existing errors (`PrimerPagination.svelte` `rel` prop, `routes/+page.svelte` `sortBy` typing) on `main` as on this branch. Zero new errors introduced.

## Out of scope

- Issue 3 (Prompt | Scenario segmented control + paired-transcript view) is M effort and needs the divergent-transcript view to ship honestly; deferred to a separate PR.
- Run-detail page (`/suite/[id]/[run_id]`) has its own column conventions — a follow-up audit can align landing and detail.

### DIFF
--- a/viewer/src/lib/PrimerDropdown.svelte
+++ b/viewer/src/lib/PrimerDropdown.svelte
@@ -7,6 +7,13 @@
   interface DropdownOption {
     value: string;
     label: string;
+    // When true, the option is rendered with reduced opacity and an optional
+    // inline `secondaryLabel` hint. Click handlers still fire so callers can
+    // implement swap-style behavior (discoverability over prevention).
+    disabled?: boolean;
+    // Optional muted, smaller-font hint shown next to the main label. Useful
+    // for explaining why an option is in a disabled visual state.
+    secondaryLabel?: string;
   }
 
   interface Props {
@@ -144,16 +151,16 @@
       tabindex={-1}
     >
       {#each options as option, idx}
-        <li class="ActionList-item" role="option" aria-selected={selected === option.value}>
+        <li class="ActionList-item" role="option" aria-selected={selected === option.value} aria-disabled={option.disabled || undefined}>
           <button
             type="button"
-            class="ActionList-content w-full text-left {selected === option.value ? 'ActionList-content--selected' : ''} {highlightedIndex === idx ? 'ActionList-content--highlighted' : ''}"
+            class="ActionList-content w-full text-left {selected === option.value ? 'ActionList-content--selected' : ''} {highlightedIndex === idx ? 'ActionList-content--highlighted' : ''} {option.disabled ? 'ActionList-content--disabled' : ''}"
             onclick={() => handleSelect(option.value)}
             onmouseenter={() => (highlightedIndex = idx)}
             onmouseleave={() => (highlightedIndex = -1)}
           >
             <span class="ActionList-content-text">
-              {option.label}
+              {option.label}{#if option.secondaryLabel}<span class="ActionList-content-secondary">{option.secondaryLabel}</span>{/if}
             </span>
           </button>
         </li>
@@ -295,6 +302,17 @@
     overflow: hidden;
     text-overflow: ellipsis;
     white-space: nowrap;
+  }
+
+  :global(.ActionList-content--disabled) {
+    opacity: 0.5;
+  }
+
+  :global(.ActionList-content-secondary) {
+    margin-left: 0.5rem;
+    color: var(--fgColor-muted, #59636e);
+    font-size: 0.75rem;
+    font-weight: 400;
   }
 
   :global(html[data-color-mode="dark"] .ActionMenu-button) {
--- a/viewer/src/lib/compare-view.ts
+++ b/viewer/src/lib/compare-view.ts
@@ -14,23 +14,31 @@ export function buildMatchedSampleRows(
 	runIds: string[],
 	metric: string,
 	disagreementsOnly: boolean,
-	baselineRunId?: string
+	baselineRunId?: string,
+	// Optional: customize the key used to match samples across runs. Defaults to
+	// matching by `sample.prompt`, which is appropriate for single-turn prompts
+	// comparisons. Scenario comparisons pass `(s) => s.test_case_id ?? s.prompt`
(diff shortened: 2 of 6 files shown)


changliu2:
Added commit `a2658eb` — extends the populated-dim column logic to also surface **custom per-suite dim names** in the suite-landing `Evaluation results` table.

The fix in 7dea023 derived headers from `data.dimensionDefs` and filtered out empty columns, but `dimensionDefs` only contains `BUILT_IN_DIMENSIONS` (`policy_violation` / `overrefusal` / `harm_actionability`) plus the global `examples/eval-definitions/judge_dimensions.yaml` file. Custom dims declared inline in a suite's `eval.yaml` `judge.dimensions` block (e.g. `safety_violation` and `unjustified_refusal` in the bank-manager-agent-shield demo) never reach `loadDimensions()`, so they were silently dropped from the suite-landing table even though they appear in every run's `metrics.dimensions`.

Now derives the dim-name set as the union of (a) `dimensionDefs` keys and (b) `Object.keys(run.{prompt,audit}.metrics.dimensions)` across all runs — mirroring the pattern `loadComparePageData` already uses (`allMetrics = union of Object.keys(summary.dimensions)`). `visibleDimNames` keeps the existing filter that drops dims with no data, so `harm_actionability` still hides for evals that don't score it.

Verified locally against `artifacts/results/bank-manager-agent-shield/` — both `variant-a-unguarded-n100` and `variant-e-guarded-acs-n100` write all four dim keys to `metrics.dimensions` (e.g. `safety_violation: {rate: 0.22, count: 50, ...}` in the variant-a prompt run), so the table now picks them up and renders `Safety violation` / `Unjustified refusal` columns alongside `Policy violation` / `Overrefusal`.

`npm run check` reports the same 3 pre-existing errors documented in the PR body. Zero new errors.

changliu2:
Stacked 2 more commits on this PR addressing follow-up evaluation-results table UX issues:

1. **afd95bf** — `refactor(viewer/behavior-table): move target into run name cell`. The target string was being rendered once per subrow (aggregate row + `Prompts` subrow + `Scenarios` subrow), repeating the same value 3x within every run group. Dropped the standalone `Target` column and surface the target as a secondary monospace line under the run name link. Frees one fixed column slot for dim columns and removes ~30% of horizontal noise on demos with long target identifiers like `examples.bank_manager_agent_shield.agent:chat_guarded_acs`.

2. **d74e009** — `feat(viewer/behavior-table): dimension columns are now dropdowns`. Each of the up-to-3 metric column headers is now a `<select>` populated from `visibleDimNames` (the same data source the populated-dims fix in a2658eb already derives). Default = first 3 populated dims in their existing order, so the table renders identically to before on first load. The selection persists in the URL search param `?metrics=<comma-list>` for shareable links. Selecting a dim already shown in another column swaps the two slots so the same dim is never duplicated; options for already-used dims are also disabled in other dropdowns as a visual hint. The `Total` column is unchanged. Em-dash `—` placeholder logic from a2658eb is preserved when a run doesn't have the chosen dim.

CI: `npm --prefix viewer run check` reports the same 3 errors / 6 warnings baseline as this PR's prior commits — zero new diagnostics in the edited file.

tangym (approved):

PR #181, opened 2026-06-01
docs+examples: post-rename cleanup, jargon scrub, broken nav fixes

changliu2:
## TL;DR

Closes 14 findings from an internal customer-readiness audit of the
preview repo. All changes are scoped to docs, examples, scripts, and
test guards — **no `assert_ai/` core code changes** in this PR.

Nine logically-grouped commits (one cluster per commit) make this easy
to review file-by-file.

## What changed

- **`tests: tighten p2m guard regex`** — the previous `\bp2m\b`
  word-boundary regex missed env-var leaks like `P2M_AZURE_DEPLOYMENT`
  because the underscore broke the word boundary. Switched to a
  case-insensitive substring match, added skip lists for binary
  extensions (`.svg`) and lockfiles (`package-lock.json`, `uv.lock`,
  etc.) that occasionally embed `p2m` in sha512 hashes, and added a
  sanity test asserting the new pattern catches the historical leak.
- **`examples: rename P2M_* env vars to ASSERT_*`** — three quickstart
  READMEs still referenced `P2M_AZURE_DEPLOYMENT` / `P2M_TARGET_MODEL`
  even though the `agent.py` files already read `ASSERT_*`. Customers
  following the docs verbatim would hit a missing-deployment error.
- **`examples/docs: replace internal model names with public substitutes`** —
  swapped `azure/gpt-5.4*` for `azure/gpt-4o*` and `azure/gpt-4o-mini`
  across docs, example configs, example agent code, and the website
  snippets. `assert_ai/`, `tests/`, `.github/workflows/`, and frozen
  `artifacts/results/**` are intentionally **not** touched.
- **`AGENTS.md: complete rename to ASSERT`** — PR #178 fixed the GitHub
  URL but the prose, title, and doc cross-links still said "Adaptive
  Eval" and pointed at three moved/renamed docs (`quickstart.md`,
  `writing-eval-specs.md`, `reading-results.md`). Same fix applied to
  `.cursorrules`, `.devcontainer/devcontainer.json`, `CONTRIBUTING.md`,
  `SUPPORT.md`, and `examples/azure_doc_qa/IMPROVEMENT_JOURNEY.md`.
- **`examples/change_control_agent: genericize internal Microsoft infra references`** —
  rewrote internal Microsoft system names (Safefly, Ev2, R2D, ICM, ADO,
  ChangeKeep) to generic enterprise terminology (Deployment Gateway,
  Rollout Service, Release Readiness, incident-tracker, change-tracker,
[...]

PR #182, opened 2026-06-01
build: wheel-safe packaging, CI build workflow, env-var rename

tangym:
## Summary

Hardens packaging and the build/release surface for `assert-ai` so the wheel is publishable and installable cleanly, and standardizes the env-var prefix.

## Commits (small + ordered)

1. **`refactor(prompts):`** move `internal-pipeline-prompts/` under `assert_ai/` as a subpackage so prompt assets ship inside the wheel instead of being a sibling repo-only directory.
2. **`fix(io):`** resolve `PROMPTS_DIR` via `importlib.resources` (`files()`) so prompt lookup works whether the package is installed from a wheel, an editable install, or sdist — no reliance on `__file__` path math.
3. **`build:`** drop `examples*` from `packages.find` include so the `examples/` tree is no longer accidentally packaged into the wheel.
4. **`build(pyproject):`** flesh out package metadata — `license`, `authors`, `keywords`, `classifiers`, project URLs, and explicit `package-data` for the bundled prompt `.md` files. Required for a publishable wheel and PyPI listing.
5. **`ci:`** add `.github/workflows/build.yml` — PEP 517 build + `twine check` on every PR/push, plus a 3-OS × 3-Python (3.11/3.12/3.13) `pip install <wheel>` + `assert-ai --help` smoke matrix. Uploads `dist/` as an artifact (14d on PR, 90d on `main`).
6. **`chore(env):`** rename `ASSERT_EVAL_*` env vars to `ASSERT_AI_*` to match the package rename in #177. Includes the matching string update in `viewer/src/lib/server/run-spawn.ts` for `ASSERT_AI_COMMAND` so the viewer's spawn override stays consistent.
7. **`docs(examples):`** fix stale `P2M_*` env-var names in example READMEs so docs match the code.

## Verification

- `python -m build` → clean wheel + sdist
- `twine check dist/*` → PASSED
- Wheel contains the 12 packaged prompt `.md` files
- Smoke install + `assert-ai --help` works in a clean venv
- Guard test `tests/test_no_p2m_references.py` passes
- All commits authored as `tangym`

## Follow-up

Tracked as a separate issue: one internal TypeScript symbol rename in the viewer (`spawnAssertEvalRun`/`resolveAssertEvalCommand` → `*AssertAi*`). Kept out of this PR to avoid mixing viewer-side changes with packaging work.


changliu2 (approved):
LGTM 🚀 — clean closure of three follow-up items I had queued up for tomorrow:

- **M-2 (wheel safety):** the `importlib.resources.files()` approach for `PROMPTS_DIR` is exactly the right fix — keeps the package importable from a wheel without any path gymnastics. Moving `internal-pipeline-prompts/` under `assert_ai/internal_pipeline_prompts/` also makes the package self-contained.
- **Env-var consistency:** thanks for catching `viewer/src/lib/server/run-spawn.ts` and `assert_ai/tools.py` — those were on my list. Documentation in the 4 example READMEs also caught up.
- **CI build smoke (PEP 517 × 3-OS × 3-Python):** great addition. This is table-stakes hygiene that several best-in-class OSS projects do and we were missing.

Heads up — I have an open follow-up PR (#181) that touches some of the same 4 example READMEs (different hunks: jargon cleanup + telemetry framing). I'll rebase #181 on top of this after it merges and drop any conflicting hunks. No action needed from you.

A couple of optional follow-ups I'll pick up separately (not blocking this PR):
- CHANGELOG / README migration note for `ASSERT_EVAL_*` → `ASSERT_AI_*` so existing users find the rename
- TypeScript symbol rename in viewer: `spawnAssertEvalRun` / `resolveAssertEvalCommand` → `spawnAssertAi*` (the env-var values are right, the function names still carry the old prefix)

Merging.

PR #194, opened 2026-06-01
Website: polish docs page, add favicon

sooyeonni:
Follow-up to #179.

## Changes
- Render README.md as the `/docs` landing page (replaces card grid)
- Sidebar order: ASSERT Documentation → Getting Started → Concepts → groups
- Uppercase known acronyms (CLI, API, etc.) in auto-generated labels
- Remove breadcrumbs from individual doc pages
- Redirect legacy `/docs/readme` to `/docs`
- Add `app/icon.svg` favicon


changliu2:
Nice cleanup — promoting `docs/README.md` to the `/docs` landing reads much better, and the legacy `/docs/readme` redirect is a good catch.

One blocker before merge: `website/app/icon.svg` is a raw Figma export. The gradient is drawn via `<foreignObject>` + an HTML `conic-gradient`, and the `<path>` itself has no `fill`. Browser favicon renderers (and most non-DOM SVG consumers) don't execute `foreignObject`, so the tab icon will most likely show up blank. Could you re-export it flattened — a native SVG `<linearGradient>`/`<radialGradient>` with `fill="url(#…)"` on the circle path? Easy way to verify: open `icon.svg` directly in a browser; it should render the gradient disc, not an empty square.

Everything else LGTM. The H2-only TOC and the `ACRONYMS`/priority-sort tweaks in `_lib/docs.ts` look correct, and I confirmed the `readme.md` exclusion still allows nested section READMEs through.

minthigpen (approved):

PR #219, opened 2026-06-03
Refactor Phoenix lazy-init into assert_ai.auto_trace() helper

changliu2:
## What changed since the first review

PR #219 originally inlined a 37-line gate into `bank_manager_agent_control/agent.py`. That fixed the 30–50s `import phoenix.otel` cold-start cost but defeated the "2 lines in every agent" pitch in our README — every user would have to copy-paste 37 lines into their own `agent.py`.

This rewrite extracts the gate into `assert_ai.tracing.auto_trace()` so user code becomes:

```python
from assert_ai import auto_trace
auto_trace()
```

…which keeps the install-time pitch honest *and* fixes the slow demo startup.

## Files changed (28 total)

| Surface | Files | What changed |
|---|---|---|
| Helper (new) | `assert_ai/tracing.py` | `auto_trace(**kwargs) -> bool` + `phoenix_collector_available()`. All gate logic lives here. |
| Re-export | `assert_ai/__init__.py` | `from assert_ai import auto_trace` works. |
| Tests (new) | `tests/test_tracing.py` | 17 tests covering every gate branch + kwarg forwarding + `ImportError` handling. |
| Example agents | `bank_manager_agent_control`, `incident_triage_agent`, `change_control_agent`, `science_research_agent` | Inline `register()` blocks replaced; preserved nested-try `TracerProvider` / `_NoopTrace` fallbacks. |
| Public API | `travel_planner_langgraph/auto_trace.py`, `azure_doc_qa/auto_trace.py` | File paths preserved (referenced from yaml configs, tests, website, README); bodies simplified. |
| Gallery | 20 `phoenix_auto_trace/travel_*.py` | Uniform 2-line swap. |

## Gate behavior (unchanged from PR #219)

| Signal | Action |
|---|---|
| `PHOENIX_DISABLE_AUTO_INSTRUMENT=1` | skip |
| `PHOENIX_COLLECTOR_ENDPOINT` set | init |
| `OTEL_EXPORTER_OTLP_ENDPOINT` set | init |
| `ASSERT_AUTO_TRACE_FORCE=1` | init (bypass probe) |
| else | 100 ms TCP probe `localhost:6006`; init iff a collector is listening |

All `phoenix.otel.register()` kwargs (`project_name`, `batch`, `protocol`, `verbose`, `headers`, …) forward through.

## Validation

- `pytest tests/test_tracing.py tests/test_hosted_trace_registration.py` → **20 passed in 0.83s**
[...]

PR #220, opened 2026-06-03
fix: centralize lazy auto tracing

jakepresent:
## Summary
- Add a central `assert_ai.auto_trace` helper that installs available OpenInference instrumentors without importing Phoenix unless export is configured/reachable
- Use the helper from traced callable/hosted runtime paths so ASSERT can collect local spans without starting Phoenix export
- Replace direct `phoenix.otel.register(auto_instrument=True)` calls across examples and docs with the central helper

## Why
Importing `phoenix.otel` can pull in the Phoenix UI stack and make clean demo/CLI startup extremely slow. The helper keeps no-collector runs fast while preserving local trace capture for ASSERT's in-process collector.

## Validation
- `python3.11 -m pytest tests/test_auto_trace.py tests/test_hosted_trace_registration.py -q` -> 7 passed
- `python3.11 -m compileall -q assert_ai examples/phoenix_auto_trace examples/travel_planner_langgraph examples/azure_doc_qa examples/incident_triage_agent examples/science_research_agent examples/change_control_agent examples/travel_planner_neurosan` -> passed
- No stale direct `from phoenix.otel import register` call sites remain outside `assert_ai/auto_trace.py`
- No-collector smoke: `auto_trace.enable()` returned true from local instrumentors and did not load any `phoenix.*` modules

### DIFF
--- a/assert_ai/auto_trace.py
+++ b/assert_ai/auto_trace.py
@@ -1,0 +1,127 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Lightweight OpenInference/Phoenix auto-tracing helper.
+
+Use ``from assert_ai import auto_trace; auto_trace.enable()`` before importing
+or constructing the target agent. By default, the helper installs any available
+OpenTelemetry/OpenInference instrumentors but avoids importing ``phoenix.otel``
+when no Phoenix/OTLP collector is configured or reachable. This keeps local
+CLI/demo startup fast while still allowing ASSERT's in-process collector to
+capture spans during traced eval runs.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+import socket
+from importlib.metadata import entry_points
+
+log = logging.getLogger(__name__)
+
+_enabled = False
+_instrumentors_enabled = False
+
+
+def _can_connect(host: str, port: int, *, timeout: float) -> bool:
+    try:
+        with socket.create_connection((host, port), timeout=timeout):
+            return True
+    except OSError:
+        return False
+
+
+def _collector_available(*, timeout: float = 0.1) -> bool:
+    """Detect whether Phoenix/OTLP export is explicitly configured or local."""
+    if os.environ.get("PHOENIX_DISABLE_AUTO_INSTRUMENT") == "1":
+        return False
+
+    for name in ("PHOENIX_COLLECTOR_ENDPOINT", "OTEL_EXPORTER_OTLP_ENDPOINT"):
+        if os.environ.get(name):
+            log.info("%s is set; enabling Phoenix export", name)
+            return True
+
+    if os.environ.get("ASSERT_EXPORT_TRACES", "").strip() == "1":
+        log.info("ASSERT_EXPORT_TRACES=1; enabling Phoenix export")
+        return True
+
+    ports: list[int] = []
+    grpc_port = os.environ.get("PHOENIX_GRPC_PORT", "")
+    if grpc_port.isnumeric():
+        ports.append(int(grpc_port))
+    ports.extend([4317, 6006])
+
+    seen: set[int] = set()
+    for port in ports:
+        if port in seen:
+            continue
+        seen.add(port)
+        if _can_connect("localhost", port, timeout=timeout):
+            return True
+    return False
+
+
+def _enable_entrypoint_instrumentors() -> None:
+    """Install available OpenTelemetry instrumentors without importing Phoenix."""
+    global _instrumentors_enabled
+    if _instrumentors_enabled:
+        return
+
+    for entry_point in entry_points().select(group="opentelemetry_instrumentor"):
+        if not entry_point.value.startswith("openinference.instrumentation."):
+            continue
+        try:
+            instrumentor_cls = entry_point.load()
+            instrumentor_cls().instrument()
+        except Exception:
+            log.debug("Failed to enable OpenInference instrumentor %s", entry_point.name, exc_info=True)
+    _instrumentors_enabled = True
+
+
+def enable(
+    *,
+    auto_instrument: bool = True,
+    timeout: float = 0.1,
+    export: bool | None = None,
+    **register_kwargs: object,
+) -> bool:
+    """Enable local OpenInference tracing and optional Phoenix export.
+
+    Returns ``True`` when either local instrumentors or Phoenix registration ran.
+    Missing or misconfigured optional tracing dependencies are non-fatal so
+    examples and demos can run without Phoenix installed.
+    """
+    global _enabled
+    if _enabled:
+        return True
+
+    if os.environ.get("PHOENIX_DISABLE_AUTO_INSTRUMENT") == "1":
+        log.info("PHOENIX_DISABLE_AUTO_INSTRUMENT=1; skipping auto-tracing")
+        return False
+
+    if auto_instrument:
+        _enable_entrypoint_instrumentors()
+
+    should_export = _collector_available(timeout=timeout) if export is None else export
+    if not should_export:
+        log.info("No Phoenix/OTLP collector detected; skipping Phoenix export")
+        _enabled = _instrumentors_enabled
+        return _enabled
+
+    try:
+        from phoenix.otel import register
+    except ImportError:
+        log.info("Phoenix tracing dependencies are not installed; skipping Phoenix export")
+        _enabled = _instrumentors_enabled
+        return _enabled
+
+    try:
+        register(auto_instrument=auto_instrument, **register_kwargs)
+    except Exception:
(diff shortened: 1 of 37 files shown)


changliu2:
## Audit notes from #219 author

Disclosure: I shipped a parallel implementation in #219 (targets `build-demo-final`) hours before this. The two PRs are 80%+ overlap. Here's an honest comparison so we can pick the better one without ego getting in the way.

### Where #220 is strictly better

1. **Two-phase enable (instrumentors vs export) is the right design.** `_enable_entrypoint_instrumentors()` registering openinference instrumentors via `importlib.metadata.entry_points()` — **without** touching `phoenix.otel` — is more correct than #219, which conflates "load openinference" with "load Phoenix". Local span collection (ASSERT's in-process collector) works on a fresh laptop with no Phoenix server. #219 only side-steps the slow import; it doesn't recover instrumentation when phoenix is uninstalled but a collector endpoint exists.

2. **`export=True/False` explicit override** lets callers force-disable export even when a collector is reachable. #219 has `force=True` (force-enable) but no force-disable. The new `otel_session.py` call site `auto_trace.enable(export=False)` is a clean use of this — it instruments locally without trying to register a Phoenix exporter at all.

3. **Probes both 4317 (OTLP/gRPC) and 6006 (Phoenix UI).** #219 only probes 6006. 4317 is the actual collector port for non-Phoenix OTel stacks; this is the more correct default.

4. **Module idempotency at module scope** (`_enabled`, `_instrumentors_enabled`) protects re-entry. #219 makes the callable idempotent inside the function but the module-level guard is cleaner against test-ordering issues.

5. **Wired into `_ensure_hosted_trace_instrumentation` in `stages/inference.py`**. This was on my follow-up list and not in #219.

### Where I think #219's choices were better (worth porting)

1. **`auto_trace()` callable vs `auto_trace.enable()`**. The README pitch is "*2-line install*: `from assert_ai import auto_trace; auto_trace()`". `auto_trace.enable()` reads as more methods-to-learn for users. Trivial to fix: keep the module, add `__call__` via a module wrapper, or rename helper to e.g. `enable_auto_trace`.

2. **`ASSERT_AUTO_TRACE_FORCE=1` env var**. #220 has `ASSERT_EXPORT_TRACES=1` which only forces export. #219 also has a force-instrument escape hatch for users with non-standard collectors on unusual ports. Low cost to add.

3. **Bank-manager demo (`examples/bank_manager_agent_control/agent.py`)** is the loudest cold-start sufferer (37-line inline gate before, was the original motivation). #220 doesn't touch it because it targets `main`, but post-merge `main → build-demo-final` will need this file migrated too — or there'll be drift between the helper users and the demo speakers.

4. **`tests/test_no_p2m_references.py`** is failing on `main` and `build-demo-final` (pre-existing — flags `bank_manager_agent_control/results/.../inference_set.jsonl:86`). Not yours to fix, but worth knowing about before merge if CI gating fires.

### Conflict outlook with #219

37 overlapping files — every `phoenix_auto_trace/travel_*.py`, `examples/{change_control,incident_triage,science_research}_agent/agent.py`, `examples/{travel_planner_langgraph,azure_doc_qa}/auto_trace.py`.

Different base branches (`main` vs `build-demo-final`), so they won't auto-conflict in GitHub's UI, but whichever lands second will need a rebase. Practical paths:

- **If #220 merges first**: I close #219, port the bank-manager `agent.py` migration + README + yaml `-v2` changes as a follow-up against `main`, then cherry-pick to `build-demo-final` so the demo branch picks up #220's `auto_trace.py`.
- **If #219 merges first** (post-demo): you rebase #220 onto the new helper. Most diff churn will be rewriting `tracing.py` → `auto_trace.py` and the call-site updates.

My honest recommendation: **merge #220 to main**, I close #219, and I port the bank-manager-specific work (agent.py inline-gate removal + README + `-v2` yamls) as a small follow-up on top.

cc @changliu2_microsoft

changliu2:
## External-surface coverage audit

Followup to my [first comment](https://github.com/responsibleai/ASSERT/pull/220#issuecomment-4614642386). Question on the table: does #220 touch every external-facing sample including docs and the GitHub Pages site?

**Method:** grepped `main` for `phoenix.otel.register`, `from phoenix.otel`, `PHOENIX_*`, `auto_trace`, `two lines` across `examples/`, `docs/`, `website/`, and root `README.md`. Cross-referenced against the 37 files in #220's diff.

### Covered (all good)

| Surface | Status |
|---|---|
| 19 `examples/phoenix_auto_trace/travel_*.py` gallery files | ✅ migrated |
| 3 standalone `agent.py` (incident_triage, change_control, science_research) | ✅ migrated, fallback try-blocks preserved |
| 2 standalone `auto_trace.py` wrappers (travel_planner_langgraph, azure_doc_qa) | ✅ migrated |
| `docs/targets/README.md` and `docs/targets/callable.md` | ✅ now show `from assert_ai import auto_trace; auto_trace.enable()` |
| `examples/README.md`, `examples/phoenix_auto_trace/README.md` | ✅ updated |
| `examples/travel_planner_langgraph/README.md` ASCII diagram | ✅ updated (`auto_trace.enable()` in the flow) |
| `examples/travel_planner_neurosan/README.md` | ✅ updated |
| `pyproject.toml` comment about `register(auto_instrument=True)` | ✅ updated |
| `assert_ai/stages/inference.py` (hosted-session trace registration) | ✅ wired into helper |
| `assert_ai/core/otel_session.py` (live OTel session) | ✅ wired into helper with `export=False` |

### Real gaps (worth fixing before merge)

**1. `examples/phoenix_auto_trace/travel_langgraph.py` — the missing gallery file.**

The other 19 `travel_*.py` files in the gallery all had `from phoenix.otel import register; register(auto_instrument=True)` at the top and are now migrated to `auto_trace.enable()`. `travel_langgraph.py` has **neither** — no `register()` and no helper call. On `main` today this file imports `phoenix_auto_trace._tools` but never registers any instrumentation, so unless it's relying on `eval_config.yaml` doing it externally, it's broken/silent in the gallery. Either:
- Add `from assert_ai import auto_trace; auto_trace.enable()` at the top (consistent with the other 19)
- Or call out explicitly why it's intentionally different

**2. `docs/getting-started.md` — the flagship onboarding page.**

Tells users to install with `pip install -e ".[otel,langgraph]"` then `phoenix serve` then `assert-ai run ...`. Doesn't mention `auto_trace.enable()` anywhere. A customer landing here would never learn the new helper exists — they'd see Phoenix mentioned and assume they need `phoenix serve` running for instrumentation to work. With the new helper, that's no longer true (local-only instrumentors via entry_points cover the OTel-traced eval path even without a Phoenix server).

Suggest a short note in the "Run your first evaluation" section explaining: instrumentors install lazily via `auto_trace.enable()`; `phoenix serve` is now genuinely optional, only needed if you want a UI to browse traces. This is the user-facing payoff of your refactor and the doc shouldn't bury it.

**3. Root `README.md` L40 (low priority, deferable).**

Says "Phoenix/OpenInference auto-instruments 33+ frameworks in two lines". The "two lines" claim is now `from assert_ai import auto_trace; auto_trace.enable()` instead of `from phoenix.otel import register; register(auto_instrument=True)`. The text stays accurate either way (still two lines), but if you want to be precise the auto_trace helper is the new canonical pitch. Deferable to a docs follow-up.

### NOT gaps (intentionally skipped — flagging so you don't waste time)

- `examples/travel_planner_langgraph/agent.py`, `examples/phoenix_auto_trace/travel_langgraph.py` agent imports — only import `phoenix_auto_trace._tools` (tools module name collision; not instrumentation)
- `examples/travel_planner_neurosan/agent.py` — uses manual OTel SDK by design, intentionally framework-free
- `assert_ai/core/otel.py`, `assert_ai/core/collector.py` — internal modules
- `docs/config/schema.md` — only documents `backend: phoenix` config field
- `docs/guides/troubleshooting.md` — only mentions example name in a path reference
- `website/app/CodeSnippets.tsx` — yaml only; the `callable: ...auto_trace:chat_sync` reference points at the file you migrate internally, so the snippet stays valid
- `website/app/PartnerQuotes.tsx` — Arize partner quote mentioning Phoenix as a product, not an instrumentation pattern
- `tests/comparison/*.yaml`, `tests/regression/*.yaml`, all `examples/**/eval_config.yaml`, `examples/phoenix_auto_trace/eval_*.yaml` — config references to module paths, not instrumentation patterns

### Recommendation

Fix gap 1 (one-line edit, consistency with the other 19 gallery files) and gap 2 (short paragraph in getting-started.md) before merge. Gap 3 is deferable.

After those land, the PR is a clean "every place the user could land + see the 2-line pitch now shows the helper". Otherwise customers reading getting-started.md will keep installing+running Phoenix unnecessarily.

cc @changliu2_microsoft

changliu2:
Superseded by #224 (centralized auto_trace + customer-facing sweep), which has merged. Thanks Jake!

PR #230, opened 2026-06-05
chore: add maintainer-assist agent pattern (dev-maintainer + designer)

changliu2:
## What

Adds a reusable OSS maintainer-assist pattern: 2 Copilot CLI agents + 2 skills + observation-mode governance to keep PRs reviewed when the maintainer can't review every PR within hours. Designed to address the failure mode behind PR #214 (sat ~67h with zero reviewers requested).

## Merge order (depends on #229)

**This PR must merge after [#229 (CODEOWNERS)](https://github.com/responsibleai/ASSERT/pull/229).** The dev-maintainer reviewer-routing logic reads `.github/CODEOWNERS`, which doesn't exist on `main` until #229 lands.

## Architecture (3 layers)

| Layer | Mechanism | What it does |
|---|---|---|
| 1. Mechanical routing | `.github/CODEOWNERS` (PR #229) | Auto-requests reviewers when PR opens. |
| 2. Agent assist | Dev-maintainer agent (this PR) | When PR sits unreviewed 24h, requests a co-owner. When stale 72h, requests a second co-owner. |
| 3. Human approval | Repository maintainer | Last-resort fallback after all other owners exhausted. |

## Observation mode (default state)

Observation mode is the default and currently only enabled state.

| Agent | Default state |
|---|---|
| Dev-maintainer | Observation-only **+ 2 narrow write exceptions** (see below) |
| Designer | Strictly observation-only |

The two narrow write exceptions for the dev-maintainer are **active by default on merge** under `AGENTS.md` governance (they exist precisely to keep PRs moving):

| Write | Allowed | Forbidden |
|---|---|---|
| Audit-only PR comment | ✅ | Approving review, request-changes review |
| Request review from existing CODEOWNERS | ✅ | Merging, labelling, closing |

Any *future* broader write capability for the dev-maintainer (approve, merge, label, etc.) and any write capability for the designer agent require explicit per-agent activation by the maintainer.

## Reviewer routing logic (dev-maintainer)

When picking who to ping:

1. Exclude PR author.
2. Exclude any owner whose GitHub user status is set to "busy" / "out of office" (agent reads the GraphQL `user.status` field; owners keep this in sync via their GitHub profile).
3. Exclude the fallback admin **unless all other co-owners are excluded** — fallback admin is last-resort.
4. From remaining, prefer the owner pinged least recently for that path.

## 24-hour escalation rule

| PR age (no review action) | Agent action |
|---|---|
| < 24h | Log to dev-inbox only. |
[...]

PR #228, opened 2026-06-05
fix(web): explicit cursor:pointer on .nav-btn

changliu2:
## What

One-line CSS: add `cursor: pointer` to `.nav-btn`.

## Why

Top-nav buttons (`Documentation`, `Read the blog`, `Get started`) relied on the browser's `a:any-link` default to render the pointer cursor on hover. With `display: inline-flex` on the anchor and nested spans inside `.nav-btn-shine` (`shine-border` + `shine-label`), that browser default did not apply consistently across browsers.

User-reported repro: hovering `Read the blog` on any `/docs/*` sub-page shows the default arrow cursor, making the button look unclickable even though the click works.

## Verification

- HTML for the anchor is correct: `<a href="https://commandline.microsoft.com/assert-written-intent-executable-evals/" target="_blank" rel="noopener noreferrer" class="nav-btn nav-btn-secondary">Read the blog</a>`
- URL returns 200 OK (real published blog post)
- Existing `.nav-btn-secondary:hover` styles are unaffected
- No other site rule sets `cursor` on `a` or `.nav-btn` (verified via `grep cursor website/app/globals.css`)

Diff: 1 file, +1 / −0.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

PR #226, opened 2026-06-04
chore: pre-launch demo polish (quieter logs + faster default concurrency)

changliu2:
Pre-launch polish for the canonical demo experience. Two small commits:

## Commit 1 — Demote per-run setup chatter to DEBUG

Every `assert-ai run` of an example with `assert_ai.auto_trace` was emitting 2 INFO lines on every invocation even when nothing was wrong:

- `INFO No Phoenix/OTLP collector detected; skipping Phoenix export` (`auto_trace.py`)
- `INFO No existing TracerProvider — creating a minimal one for ASSERT` (`otel.py`)

These describe expected steady-state behavior when Phoenix is not running. They belong at DEBUG, not INFO.

Demoted to DEBUG (8 lines total, `log.info` → `log.debug`):
- `assert_ai/auto_trace.py` lines 42, 46, 100, 108, 115 (Phoenix endpoint detection, opt-in flag, opt-out flag, no-collector, missing dep)
- `assert_ai/core/otel.py` lines 601, 607, 611 (TracerProvider attach paths)

WARNING-level failure paths (e.g. `Failed to enable Phoenix auto-tracing`) intentionally left untouched.

## Commit 2 — Bump flagship example concurrency from 1 to 4

`examples/travel_planner_langgraph/eval_config.yaml` defaulted to `inference.concurrency: 1`. For the canonical demo this makes the run feel sluggish (5 prompts × 5 scenarios = 10 sequential inferences).

Measured E2E with concurrency=8: 308.5s wall, 30 inference calls, 0 rate-limit warnings against Azure tenant. Concurrency=4 leaves 4× headroom on TPM while still cutting wall time roughly in half vs the previous default. CLI override (`assert-ai run --concurrency N`) and env var (`ASSERT_AI_RUN_CONCURRENCY`) are unchanged.

## Validation

- `tests/test_auto_trace.py` — 4 passed
- E2E run on the same example, post-changes: clean output, no Phoenix boilerplate, 10/10 scored, judge failure rate 0%
- 25-sample sweep of all `auto_trace.enable()` call sites on this branch: 0/25 emit noise (was 0/25 on main pre-fix because the chatter only appears with full `assert-ai run`, not bare imports)

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

PR #88, opened 2026-05-22
examples(bank-manager): ACS-vs-unguarded comparison demo

changliu2:
# examples(bank-manager): ACS-vs-unguarded comparison demo

Self-contained ASSERT example comparing an unguarded LangGraph bank-manager agent against the same agent guarded by the [Agent Control Specification (ACS)](https://github.com/responsibleai/AgentControlSpecification) runtime on a 100-case eval suite.

## Headline result (n=100, same test set)

| variant | safety_violation | unjustified_refusal |
|---|---:|---:|
| unguarded baseline | 39% | 2% |
| **ACS-guarded** | **9%** | **2%** |

ACS cuts safety failures by ~77% with no helpfulness cost.

## What's in this PR

Everything is under `examples/bank_manager_agent_shield/`:

- `agent.py` — two ASSERT callable targets: `chat_unguarded` and `chat_guarded_acs`, both over the same LangGraph ReAct agent talking to a mock MCP banking server.
- `mcp_server.py` — mock banking MCP server (`read_account`, `read_transaction_history`, `prepare_transfer`, `request_customer_approval`, `create_transfer`, `freeze_account`, `enable_admin_mode`).
- `acs/manifest.yaml` — ACS manifest binding the Rego policy to the `input`, `pre_tool_call`, `post_tool_call`, and `output` intervention points.
- `acs/policy/bank_manager.rego` — stateless deterministic policy: SSN regex on user input, sensitivity-scoped (ACC-1002 / ACC-1003) read + transfer gates, approval / admin-mode gates over host snapshot state, generic prompt-injection scrubber on tool results.
- `eval_unguarded_v2.yaml` — baseline config; owns the full pipeline (systematize → test_set → inference → judge); 50 prompt + 50 scenario cases.
- `eval_guarded_acs.yaml` — ACS variant config; reuses the baseline suite-root `test_set` so both variants are scored against identical cases.
- `results/` — frozen n=100 artifacts for both variants (`variant-a-unguarded-n100`, `variant-e-guarded-acs-n100`) so the viewer renders the comparison without re-running the suite.
- Plus a one-row entry in `examples/README.md` pointing at the new demo.

## How to view the results

The committed `results/` tree is everything the viewer needs — no Python install required beyond the viewer dev server.

```powershell
# Windows
New-Item -ItemType Directory artifacts/results -Force | Out-Null
Copy-Item examples/bank_manager_agent_shield/results `
          artifacts/results/bank-manager-agent-shield -Recurse
[...]

PR #232, opened 2026-06-09
feat: scheduled review-escalation Action + always-on-host docs (#230 follow-up)

changliu2:
## What

Follow-up to @tangym's review note on #230: the dev-maintainer escalation windows (24h / 72h / 7d) are **wall-clock thresholds**, but the loop was documented as running on the maintainer's **local workstation** — which is offline in exactly the scenario the escalation exists for (the maintainer being away). This PR (1) fixes the docs to require an always-on host and (2) ships that host as a scheduled GitHub Action so the escalation actually fires unattended.

> _"the escalation logic runs in a local CLI loop on the maintainer's workstation. If the maintainer is unavailable, the laptop likely is too. So the 24h/72h rules won't fire in exactly the scenario that triggered this."_ — #230 review

## Changes

**Docs** (`AGENTS.md`, `.github/agents/dev-maintainer.md`)
- New **"Where to run the loop"** section: run on an always-on host (small VM, CI cron, or scheduled GitHub Action) — never the maintainer's laptop — and a note that `.github/CODEOWNERS` already covers the baseline independent of the loop.
- The pattern intro no longer says "local workstation."

**Reference implementation** (`.github/workflows/review-escalation.yml` + `.github/scripts/escalate_reviews.py`)
- Scheduled workflow (every 6h) + `workflow_dispatch` (with a **dry-run default** for safe manual runs).
- `escalate_reviews.py` is the **deterministic** half of the dev-maintainer agent — pure CODEOWNERS routing, **no LLM** — so it runs reliably in CI. The LLM `audit-pr` pass stays a separate concern a maintainer runs from any host.
- Routing implements the `AGENTS.md` / `dev-maintainer.md` contract: last-match CODEOWNERS → exclude the PR author → exclude OOO owners (best-effort via GitHub user status) → fallback admin only as last resort, **never the author**. Deterministic tie-break (now documented in both files): most changed-path coverage, then alphabetical — the stateless Action uses this in place of 'least recently pinged'. Cascade is evaluated by severity (7d → 72h → 24h) so the 7-day fallback is always reachable.
- `permissions: { contents: read, pull-requests: write }`; uses the built-in `GITHUB_TOKEN`.

## Test plan

Validated by running the exact script in `--dry-run` against **all open PRs** (a `workflow_dispatch` on a brand-new workflow can't be triggered until it's on the default branch, so the script is locally runnable for this reason):

- **#231** (this stack's sibling, fresh, owners already requested) → `observe`; routing preview ranks the correct owners: `@jakepresent, @AaronAspinwall123, @tangym, @minthigpen, @sooyeonni`, author `@changliu2` excluded. No double-ping.
- **#228** (113h, no reviewer) → would request `@minthigpen`; **#226 / #206** → `@AaronAspinwall123`.
- PRs that already have a reviewer → `observe`; authors always excluded; fallback admin only when they are the sole owner.

YAML well-formed, script `py_compile`s clean.

> Nothing fires until this merges to the default branch (both `schedule` and `workflow_dispatch` require the workflow on the default branch). After merge, the scheduled runs are live; manual `workflow_dispatch` defaults to dry-run.

### DIFF
--- a/.github/agents/dev-maintainer.md
+++ b/.github/agents/dev-maintainer.md
@@ -31,7 +31,7 @@ Watches the `responsibleai/ASSERT` repository for new pull requests and issues. 
 
 ## Observation-mode write workflow
 
-The agent runs on a recurring observation loop. For each open PR on every pass:
+The agent runs on a recurring observation loop on an always-on host (not the maintainer's workstation — the wall-clock escalation windows below only fire if the loop stays up while the maintainer is away; see the "Where to run the loop" section in `AGENTS.md`). For each open PR on every pass:
 
 1. **Run `audit-pr`** and log the result to `dev-inbox.md`. This always happens.
 2. **Check reviewer state.** If the PR has no reviewer requested, or a reviewer has not responded within the escalation windows below, the agent issues exactly one of the two permitted writes:
--- a/.github/agents/dev-maintainer.md
+++ b/.github/agents/dev-maintainer.md
@@ -54,8 +54,10 @@ Read [`.github/CODEOWNERS`](../CODEOWNERS) for the path-to-owner mapping. Then:
 
 1. **Exclude the PR author.**
 2. **Exclude any owner whose GitHub user status is set to "busy" / "out of office"** at the time the agent runs. The agent queries the GraphQL `user.status` field for each candidate; owners keep this in sync themselves via their GitHub profile.
-3. **Exclude the fallback admin** unless every other eligible owner has been excluded by the rules above. The fallback admin is the reviewer of last resort.
-4. From the remaining candidates, prefer the owner who has been pinged least recently for this path.
+3. **Exclude the fallback admin** unless every other eligible owner has been excluded by the rules above. The fallback admin is the reviewer of last resort. **Never request the PR author**: if a path's only owner is the author (e.g. the catch-all owner opened the PR), make no request and flag the PR for manual escalation instead.
+4. Pick deterministically from the remaining candidates: the owner covering the most changed paths, then alphabetical order. The reference Action (`../workflows/review-escalation.yml`, via `../scripts/escalate_reviews.py`) is **stateless**, so it uses this deterministic order rather than tracking per-path ping history; a stateful host may substitute "least recently pinged for this path." For the 72h second-owner and 7d fallback steps, already-requested owners are excluded and the next is chosen by the same order.
+
+The escalation cascade is evaluated by severity (7d → 72h → 24h) so the 7-day fallback is always reachable for a requested-but-silent PR, and the fallback step is skipped (logged as a manual escalation) whenever it would otherwise target the author.
 
 ### What this agent never does (even in observation mode)
 
--- a/.github/scripts/escalate_reviews.py
+++ b/.github/scripts/escalate_reviews.py
@@ -159,6 +159,21 @@ def _rank_owners(candidates_by_file: dict[str, list[str]]) -> list[str]:
     return sorted(coverage, key=lambda o: (-coverage[o], o))
 
 
+def _safe_fallback(author: str, requested: set[str], reviewed_by: set[str]) -> str | None:
+    """The fallback admin, or None when requesting them would be wrong.
+
+    Honors routing rule #1 (never request the PR author): if the fallback admin
+    *is* the author, or has already been requested or has already reviewed,
+    return None so the caller escalates for manual handling instead of pinging
+    the author / re-pinging the same person.
+    """
+    if FALLBACK_LOGIN == author:
+        return None
+    if FALLBACK_LOGIN in requested or FALLBACK_LOGIN in reviewed_by:
+        return None
+    return FALLBACK_LOGIN
+
+
 def evaluate_pr(
     repo: str,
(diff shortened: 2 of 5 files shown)


jakepresent (changes requested):
Requesting changes because the always-on host idea is right, but the deterministic escalation script does not yet match the governance contract it says it implements.

Blocking issues:

1. **The routing policy differs from `AGENTS.md` / `dev-maintainer.md`.** The docs say to exclude the author, exclude OOO users, keep fallback admin as last resort, then prefer admins / the owner pinged least recently. The script instead ranks by changed-file coverage and alphabetic tie-break, with no admin preference and no ping-history state. That may be a reasonable deterministic policy, but then the source-of-truth docs need to say that. Right now the PR body says the script mirrors `AGENTS.md`, and it does not.

2. **The 7-day fallback path is effectively unreachable for the normal stale-request case.** In `evaluate_pr`, the `requested and not has_reviewed and age >= WINDOW_72H` branch runs before the `age >= WINDOW_7D` branch. So a PR with requested reviewers and no response at 7+ days either keeps adding another non-fallback owner every run, or if no additional owner exists, returns “72h reached but no additional eligible owner” with no fallback request. That contradicts the documented “≥ 7 days, still no response -> fallback admin” rule.

3. **The fallback branch can request the PR author.** The candidate pool excludes the author, but if the pool is empty the code sets `chosen = [FALLBACK_LOGIN]` anyway. For a PR authored by the fallback admin that only matches the catch-all owner, this would try to request review from the author, violating the first routing rule. It should no-op / log an anomaly / require manual escalation instead.

I verified the script `py_compile`s and a dry run works, so this is not a syntax/build issue. It is specifically the escalation semantics versus the governance text.

changliu2:
Thanks @jakepresent — all three are real and all three are fixed in `e650e32`. You correctly identified that the script's escalation semantics diverged from the governance text it claims to implement.

**1. Routing now matches the governance docs (and vice-versa).**

You're right: the script ranked by changed-file coverage + alphabetical, while `AGENTS.md`/`dev-maintainer.md` said "exclude author → exclude OOO → fallback last resort → prefer admins / least-recently-pinged." The stateless Action genuinely can't do "least recently pinged for this path" without cross-run state, so rather than fake it I made the docs state the deterministic policy the reference impl actually uses, and kept the exclusion rules as the contract:

| Surface | Change |
|---|---|
| `AGENTS.md` §Reviewer routing | Step 5 rewritten: deterministic tie-break = most changed-path coverage, then alphabetical; explicit note that the stateless Action uses this *in place of* "least recently pinged" (a stateful host may substitute it). Added the never-request-the-author rule. |
| `.github/agents/dev-maintainer.md` §Reviewer routing | Same tie-break wording + author-guard; added a note that the cascade is severity-ordered (7d → 72h → 24h) so the 7d fallback is reachable. |
| PR body | "mirrors AGENTS.md" → precise description of the implemented contract. |

If you'd prefer the script instead grow real least-recently-pinged behavior (derivable from the PR timeline's `review_requested` events), I'm happy to do that as a follow-up — but I didn't want to claim a fairness property the stateless Action doesn't actually have.

**2. The 7-day fallback is now reachable.**

Confirmed bug — the old `requested and age >= 72H` branch shadowed the `age >= 7D` branch. `evaluate_pr` now cascades by severity:

```
reviewed already            → observe
never requested             → request first eligible owner (any age past the window)
requested + ≥ 7d, silent    → fallback admin (last resort)
requested + ≥ 72h, silent   → second NON-fallback owner (fallback reserved for 7d)
requested + < 72h           → observe (within window)
```

So a requested-but-silent PR at 7d+ now hits the fallback branch. I also tightened 72h to use **non-fallback** owners only, so the fallback admin isn't pulled in early.

**3. The fallback can no longer request the PR author.**

New `_safe_fallback()` returns `None` when the fallback admin is the author (or has already been requested/reviewed). The caller then widens to another owner, or — if the author is genuinely the only owner of the path — emits `action="manual"` with a `::warning::` for human escalation, instead of pinging the author.

Verified in a dry-run across all open PRs: **PR #88 (authored by `@changliu2`, the catch-all owner) now routes to `@AaronAspinwall123`, not the author.** No open PR targets its own author.

**Tests:** added `tests/test_escalate_reviews.py` (7 cases) covering all three regressions — 7d-fallback-reachable, fallback-never-requests-author, 72h-uses-non-fallback-owner — plus author-exclusion on first request and the observe paths. All green; `py_compile` clean; full dry-run re-run.

Re-requesting review when you have a sec. Genuinely good catch on the semantics-vs-docs gap. 🙏

jakepresent (approved):
Re-reviewed after the follow-up fixes. The blockers from my previous review are addressed:

- The docs now describe the deterministic stateless routing policy instead of claiming ping-history behavior.
- The 7-day fallback branch is reachable before the 72h branch for requested-but-silent PRs.
- The fallback path now avoids requesting the PR author and logs/manual-escalates when fallback would be invalid.

Validation:

- `python -m pytest tests/test_escalate_reviews.py -q` -> `7 passed`
- `python -m py_compile .github/scripts/escalate_reviews.py` passed

Looks good to me.

PR #233, opened 2026-06-09
chore: drop 19 MB of committed per-case scores.jsonl from incident-triage demo

changliu2:
## What

The four incident-triage variant runs committed their full per-test-case judge outputs (\scores.jsonl\, ~5 MB each = **19.1 MB total**), bloating every clone and CI checkout. This untracks them and re-ignores the pattern.

## Kept (the README + viewer rely on these)

- \metrics.json\ — the aggregate numbers cited in the README results table
- \manifest.json\, \config.yaml\ — per-variant run metadata
- \	rade_off.png\ — the chart the README embeds (unchanged)

\scripts/render_trade_off.py\ already falls back to the documented case-study numbers when \scores.jsonl\ is absent (see its module docstring), so the chart still re-renders.

## Tradeoff

Viewer **per-case drill-down** on these four pre-computed demo runs now shows summary metrics only (no per-row judge data). Re-run the suite locally to regenerate \scores.jsonl\ (gitignored going forward).

## Scope

- 4 files deleted, 19.1 MB
- \.gitignore\: re-ignore \xamples/incident_triage_agent/artifacts/**/scores.jsonl\ (the explicit artifact negation above it still allows the small curated files)
- Local working copies preserved (\git rm --cached\ only)

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,11 @@ artifacts/
 # so the README can reference it before any live n=200 run.
 !examples/incident_triage_agent/artifacts/
 !examples/incident_triage_agent/artifacts/**
+# ...but never the per-test-case judge rows: ~5 MB each, regenerated on every
+# run. The curated summaries (metrics.json, manifest.json, config.yaml) and the
+# trade_off.png chart are kept; render_trade_off.py falls back to the documented
+# case-study numbers when scores.jsonl is absent.
+examples/incident_triage_agent/artifacts/**/scores.jsonl
 
 # Do not ignore SvelteKit route folders named "logs"
 !viewer/src/routes/**/logs/
--- a/examples/incident_triage_agent/README.md
+++ b/examples/incident_triage_agent/README.md
@@ -394,7 +394,7 @@ story is the A → C demo path: ACS collapses the procedural axis at
 a visible overrefusal cost.
 
 **Artifacts.** Config: `eval_config_naive_prompt.yaml`. Results:
-`artifacts/results/incident-triage-agent-v1/naive-prompt/{config.yaml,manifest.json,metrics.json,scores.jsonl}`.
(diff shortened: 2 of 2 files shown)


jakepresent (dismissed):
Approved. I verified the ignore rule behaves the way the PR intends: the generated incident-triage `scores.jsonl` files are ignored, while the curated `metrics.json`, `manifest.json`, `config.yaml`, and `trade_off.png` artifacts remain trackable. I also checked there are no tracked incident-triage `scores.jsonl` files left and ran `python scripts/render_trade_off.py --suite incident-triage-agent-v1 --out /tmp/incident-triage-tradeoff-pr233.png` to confirm the fallback path still renders without the committed score rows.

One non-blocking follow-up worth considering later: the incident-triage README still mentions `scores.jsonl` in a couple of artifact path summaries. That is still true for freshly regenerated local runs, but less true for the checked-in demo snapshot after this cleanup.

changliu2:
Addressed your non-blocking follow-up in `bc4aede` (thanks for the thorough verify on the render fallback).

The two per-variant artifact summaries (Appendix B.1 `naive-prompt` and B.2 `guarded-with-shield-gepa`) listed `scores.jsonl` among the committed files. After this cleanup it's gitignored and regenerated locally, so both summaries now list the committed curated files — `{config.yaml, manifest.json, metrics.json}` — and note `scores.jsonl` is regenerated on a fresh run. The render-fallback mentions (the `scripts/render_trade_off.py` section) and the general "each run writes … scores.jsonl" line are left as-is since they're accurate for fresh local runs.

Docs-only change on top of your approval — re-requesting a quick re-approval per the new commit.

minthigpen (approved):

minthigpen:
Approved docs-only change on top of Jake's first approval.

PR #236, opened 2026-06-12
feat: add ACS guardrail adapter to ASSERT

liamcrumm:
## Summary

Adds an adapter that turns a completed ASSERT run's findings into a deployable Agent Control Specification (ACS) policy, validates that policy against the run's known-bad examples, and lets you re-run the same target guarded. It closes the loop from "ASSERT found these failures" to "block them at runtime." Shipped as the optional `assert-ai[acs]` extra, so the base install and CLI are unaffected when it is not present.

## Motivation / linked issue

ASSERT discovers how an agent fails (per-behavior and per-dimension violation rates in `scores.jsonl`, failing behavior categories with judge reasoning, taxonomy permissibility). ACS (in the Agent Governance Toolkit) is a stateless, fail-closed runtime policy engine that can block those failures at intervention points. There was no path between the two. This adapter provides it, reusing AGT's `acs-generator` to synthesize the policy rather than hand-rolling Rego.

## Changes

New package `assert_ai/integrations/acs/`:

| Module | What it does |
|--------|--------------|
| `findings.py` | Summarizes a run (`scores.jsonl` + `inference_set.jsonl` + `taxonomy.json`) into per-behavior and per-dimension violation rates plus known-bad `FailingExample`s, reusing existing `assert_ai.results`/`assert_ai.core.judge` helpers. |
| `prompt_builder.py` | Deterministically renders findings into a natural-language guardrail prompt and tool inventory for the generator (with secret redaction). |
| `language_model.py` | Adapts ASSERT's LiteLLM model client to the ACS generator's `LanguageModel` protocol (`assert`, `openai-compatible`, `fake`). |
| `generate.py` | `generate_policy(...)` produces `manifest.yaml` + a Rego bundle + `report.md`. |
| `validate.py` | `validate_policy(...)` replays known-bad examples through the native ACS SDK and reports how many the policy handles vs. allows. |
| `guard.py` | `guard_target(...)` wraps a callable target with ACS enforcement for a guarded re-run. |

Surface and supporting changes:

- CLI group `assert-ai acs` with `generate` and `validate` subcommands (lazy-imported; `--fail-on-allow` for CI gating). Runtime guarding is exposed via the `guard_target` Python API.
- New optional extra `acs` in `pyproject.toml` (`acs-generator`, `agent-control-specification`); the package degrades gracefully with a clear install hint when the extra is absent.
- Guide `docs/guides/securing-agents-with-acs.md` (linked from `docs/targets/README.md`) and an offline runnable example `examples/acs_guardrails/` (linked from `examples/README.md`).
- `CHANGELOG.md` entry under Unreleased.

## Design notes

- Generation reuses AGT's LLM-backed `acs-generator`; review the generated Rego before deploying.
- Validation and guarding use the native runtime as the source of truth for which intervention points are enforced, so they honor `extends`-inherited points. Fail-closed `runtime_error:` verdicts are not counted as genuine policy reactions.
- The adapter centers on the `output` intervention point, where ASSERT's built-in `policy_violation` dimension scores the target response. Non-actionable dimensions (for example `overrefusal` or custom dimensions) are reported but not translated into deny rules.

## Testing

- `pytest tests/test_acs_*.py` (35 tests, covering findings extraction, prompt synthesis, the language-model factory, generation, round-trip validation, runtime guarding, and the CLI). Native-runtime tests skip gracefully when the ACS SDK or `opa` is unavailable.
- `ruff check --select E,F,W --ignore E501` clean on changed files; `python examples/acs_guardrails/demo.py` runs green offline.

## Checklist

- [x] Tests pass locally (`pytest`).
- [x] Docs updated (new guide, example, target index link, CHANGELOG).
- [x] No breaking change (additive optional extra); CHANGELOG entry included.


jakepresent (changes requested):
Thanks for pushing this direction. The ASSERT -> ACS loop is exactly the kind of integration that seems worth exploring, and I like that the base install stays unaffected behind an optional extra.

I found a few issues that I think should be fixed before merge:

1. `uv.lock` is stale for the new `acs` extra. `pyproject.toml` defines the extra, but frozen uv installs cannot see it yet. I verified `uv sync --extra acs --frozen --dry-run --no-dev --python 3.11` fails with `Extra 'acs' is not defined in the project's optional-dependencies table`. Please regenerate/update the lockfile.

2. ACS findings are grouped by the score row behavior rather than the actually violated taxonomy node. In `findings.py`, `_behavior_policy_metrics` / `_behavior_policy_metrics_from_rows` use `row_behavior(row)` / `_score_row_behavior(row)`. If a row is generated under behavior A but the judge marks behavior B as the violated `node_judgments` entry, the ACS summary/prompt attributes the failure to A. That can invert policy intent, especially when A is permissible and B is not. For policy generation, the violated taxonomy node should drive the guardrail behavior/definition/permissibility.

3. Tool-call violations are not extracted into ACS intervention points. `_example_from_row` only recovers final assistant/user `add_message` text; it ignores `tool_call`, `tool_result`, `tool_name`, and `tool_args`. I verified a transcript containing only a destructive tool call produces no failing examples. Since the prompt builder already has `pre_tool_call` / `post_tool_call` support, either this PR should populate those examples or explicitly scope the adapter to output-only guardrails for now.

4. Prompt redaction before LLM-backed policy generation is too narrow. Representative target outputs and judge rationales can include tool outputs/log text and are sent into the generator prompt. The current `_SECRET_PATTERNS` catch only a few simple forms. Please strengthen this before sending ASSERT artifacts to the policy-generation model, or otherwise make the external-LLM data boundary explicit and safer.

A couple of non-blocking but related points:

- The guide mentions OPA, but the example README only says to install `.[acs]` and run the demo. Since the Python extra cannot install the OPA binary, the example should also call out the OPA prerequisite.
- `evaluate_only` in `guard_target` still returns transformed inputs/outputs if ACS reports a transform. If that is intended ACS shadow-mode behavior, the docs should say so; otherwise shadow mode should not mutate the target path.
- `validate_policy` counts `warn` as handled, so `--fail-on-allow` can pass a policy that only warns on known-bad examples. That may be okay if validation means “reacted,” but the PR summary frames this as blocking failures at runtime, so the CLI/reporting should distinguish handled vs strongly blocked.

Validation I ran locally:

- `python -m pytest tests/test_acs_cli.py tests/test_acs_generate.py tests/test_acs_validate.py tests/test_acs_guard.py tests/test_acs_language_model.py tests/test_acs_findings.py tests/test_acs_prompt_builder.py -q` -> `12 passed, 5 skipped` in the base environment; skips were due to missing optional ACS stack / OPA.
- `python -m compileall -q assert_ai/integrations/acs examples/acs_guardrails/demo.py` passed.
- `python -m pip install -e '.[acs]' --dry-run` resolved the Python packages.
- `uv sync --extra acs --frozen --dry-run --no-dev --python 3.11` failed because `uv.lock` is stale.
- Small in-memory repros for behavior misattribution and missing tool-call extraction.

liamcrumm:
> Thanks for pushing this direction. The ASSERT -> ACS loop is exactly the kind of integration that seems worth exploring, and I like that the base install stays unaffected behind an optional extra.
> 
> I found a few issues that I think should be fixed before merge:
> 
> 1. `uv.lock` is stale for the new `acs` extra. `pyproject.toml` defines the extra, but frozen uv installs cannot see it yet. I verified `uv sync --extra acs --frozen --dry-run --no-dev --python 3.11` fails with `Extra 'acs' is not defined in the project's optional-dependencies table`. Please regenerate/update the lockfile.
> 2. ACS findings are grouped by the score row behavior rather than the actually violated taxonomy node. In `findings.py`, `_behavior_policy_metrics` / `_behavior_policy_metrics_from_rows` use `row_behavior(row)` / `_score_row_behavior(row)`. If a row is generated under behavior A but the judge marks behavior B as the violated `node_judgments` entry, the ACS summary/prompt attributes the failure to A. That can invert policy intent, especially when A is permissible and B is not. For policy generation, the violated taxonomy node should drive the guardrail behavior/definition/permissibility.
> 3. Tool-call violations are not extracted into ACS intervention points. `_example_from_row` only recovers final assistant/user `add_message` text; it ignores `tool_call`, `tool_result`, `tool_name`, and `tool_args`. I verified a transcript containing only a destructive tool call produces no failing examples. Since the prompt builder already has `pre_tool_call` / `post_tool_call` support, either this PR should populate those examples or explicitly scope the adapter to output-only guardrails for now.
> 4. Prompt redaction before LLM-backed policy generation is too narrow. Representative target outputs and judge rationales can include tool outputs/log text and are sent into the generator prompt. The current `_SECRET_PATTERNS` catch only a few simple forms. Please strengthen this before sending ASSERT artifacts to the policy-generation model, or otherwise make the external-LLM data boundary explicit and safer.
> 
> A couple of non-blocking but related points:
> 
> * The guide mentions OPA, but the example README only says to install `.[acs]` and run the demo. Since the Python extra cannot install the OPA binary, the example should also call out the OPA prerequisite.
> * `evaluate_only` in `guard_target` still returns transformed inputs/outputs if ACS reports a transform. If that is intended ACS shadow-mode behavior, the docs should say so; otherwise shadow mode should not mutate the target path.
> * `validate_policy` counts `warn` as handled, so `--fail-on-allow` can pass a policy that only warns on known-bad examples. That may be okay if validation means “reacted,” but the PR summary frames this as blocking failures at runtime, so the CLI/reporting should distinguish handled vs strongly blocked.
> 
> Validation I ran locally:
> 
> * `python -m pytest tests/test_acs_cli.py tests/test_acs_generate.py tests/test_acs_validate.py tests/test_acs_guard.py tests/test_acs_language_model.py tests/test_acs_findings.py tests/test_acs_prompt_builder.py -q` -> `12 passed, 5 skipped` in the base environment; skips were due to missing optional ACS stack / OPA.
> * `python -m compileall -q assert_ai/integrations/acs examples/acs_guardrails/demo.py` passed.
> * `python -m pip install -e '.[acs]' --dry-run` resolved the Python packages.
> * `uv sync --extra acs --frozen --dry-run --no-dev --python 3.11` failed because `uv.lock` is stale.
> * Small in-memory repros for behavior misattribution and missing tool-call extraction.

Thank you for the review. Some good catches here. I went ahead and made a revision to address. Let me know if you have any questions.

jakepresent:
Re-reviewed the updated head. The main blockers from my prior review look addressed now:

- `uv.lock` includes the `acs` extra; frozen uv dry-run now works.
- Findings are attributed to the violated `node_judgments` behavior rather than only the score row behavior.
- Tool-call violations now produce `pre_tool_call` examples.
- OPA is called out in the example README.
- `evaluate_only` no longer applies transforms to the target path.
- Warn-only policies are now distinguishable from strong blocks via `--require-block` / `fully_blocked`.

Validation I ran:

- `uv sync --extra acs --frozen --dry-run --no-dev --python 3.11`
- `python -m pytest tests/test_acs_cli.py tests/test_acs_generate.py tests/test_acs_validate.py tests/test_acs_guard.py tests/test_acs_language_model.py tests/test_acs_findings.py tests/test_acs_prompt_builder.py -q` -> `27 passed, 5 skipped`
- `python -m compileall -q assert_ai/integrations/acs examples/acs_guardrails/demo.py`
- Small in-memory repros for behavior attribution and tool-call extraction

One remaining redaction gap before I clear my changes-requested review: JSON-style secret fields are still not redacted. The current patterns catch `api_key=value` / `api_key: value`, but not quoted JSON like:

```json
{"api_key": "EXAMPLE_SECRET_VALUE"}
```

Since representative outputs and judge rationales can include JSON logs/tool output and may be sent to the policy-generation model, please add coverage for quoted JSON/YAML-ish secret fields (`"api_key": "..."`, `"password": "..."`, etc.) and a regression test. After that I expect this is re-approvable from my side.

liamcrumm:
> Re-reviewed the updated head. The main blockers from my prior review look addressed now:
> 
> * `uv.lock` includes the `acs` extra; frozen uv dry-run now works.
> * Findings are attributed to the violated `node_judgments` behavior rather than only the score row behavior.
> * Tool-call violations now produce `pre_tool_call` examples.
> * OPA is called out in the example README.
> * `evaluate_only` no longer applies transforms to the target path.
> * Warn-only policies are now distinguishable from strong blocks via `--require-block` / `fully_blocked`.
> 
> Validation I ran:
> 
> * `uv sync --extra acs --frozen --dry-run --no-dev --python 3.11`
> * `python -m pytest tests/test_acs_cli.py tests/test_acs_generate.py tests/test_acs_validate.py tests/test_acs_guard.py tests/test_acs_language_model.py tests/test_acs_findings.py tests/test_acs_prompt_builder.py -q` -> `27 passed, 5 skipped`
> * `python -m compileall -q assert_ai/integrations/acs examples/acs_guardrails/demo.py`
> * Small in-memory repros for behavior attribution and tool-call extraction
> 
> One remaining redaction gap before I clear my changes-requested review: JSON-style secret fields are still not redacted. The current patterns catch `api_key=value` / `api_key: value`, but not quoted JSON like:
> 
> ```json
> {"api_key": "EXAMPLE_SECRET_VALUE"}
> ```
> 
> Since representative outputs and judge rationales can include JSON logs/tool output and may be sent to the policy-generation model, please add coverage for quoted JSON/YAML-ish secret fields (`"api_key": "..."`, `"password": "..."`, etc.) and a regression test. After that I expect this is re-approvable from my side.

I think this is kind of going to be whack-a-mole in that you will never find a perfect redaction pattern for every conceivable way they could expose a secret. Ideally, you would expect that the llm judge would itself be inside a trusted boundary and the control of not sending secrets is more of a best-effort thing. I think in this case though it's probably the case that the raw snippet which has the potential to expose the secret isn't adding much value to the generator itself. So in this case I think I'll go ahead and remove it from the generator input.

jakepresent (approved):
This addresses the main blockers I had. The newer handling for overrefusal/permissible nodes, cited evidence turns, uncovered validation coverage, and strongly-blocked vs handled semantics all look directionally right.

I reran the focused ACS tests, compileall, and the frozen uv acs dry-run locally, and they pass. Only tiny cleanup I saw is `git diff --check` flags a blank line at EOF in `tests/test_acs_prompt_builder.py`, but nothing substantive from my side.

PR #237, opened 2026-06-13
feat(auth): Azure Managed Identity / Entra ID auth for azure/* models

tangym:
## Summary

Adds Azure Managed Identity / Entra ID authentication for `azure/*` models, alongside the existing `AZURE_API_KEY` path. Users with `azure-identity` installed can now run Adaptive Eval against Azure OpenAI without a static key — `az login`, managed identity, workload identity, etc. all work transparently via `DefaultAzureCredential`.

## How it works

Three auth modes, resolved once at process start from env vars:

| Mode | Trigger | Behavior |
|------|---------|----------|
| `key` | `AZURE_API_KEY` set | LiteLLM uses the key, no change from today |
| `aad` | `ASSERT_AZURE_USE_AAD=1` | Forces AAD; errors if `azure-identity` is missing |
| `aad-fallback` | Neither set | Uses AAD if `azure-identity` is installed, else lets LiteLLM error as before |

When AAD is active, a token provider is wired into the LiteLLM payload via `azure_ad_token_provider`. The provider wraps `DefaultAzureCredential(exclude_interactive_browser_credential=True, exclude_visual_studio_code_credential=True)` with a closure that logs the active credential class on first successful token acquisition (e.g. `AzureCliCredential (az login)`).

## Commits

- `chore: add azure-aad optional extra for managed identity auth`
- `feat: add azure_auth helper for Entra ID token providers`
- `feat: wire Azure AD token provider into LiteLLM payloads for azure/* models`
- `feat(security): redact azure_ad_token* keys in sanitize_payload`
- `docs: document Azure managed identity auth for Azure OpenAI`
- `feat(examples): add azure_managed_identity quickstart`
- `chore(logging): quiet azure-identity credential-chain INFO logs`
- `feat(auth): log resolved Azure credential after first token`
- `fix(azure-auth): defer auth-mode log emission to after configure_logging`

## Install / try it

```bash
pip install 'assert-ai[azure-aad]'
az login
unset AZURE_API_KEY
export AZURE_API_BASE=https://<your-resource>.cognitiveservices.azure.com/
export AZURE_API_VERSION=2025-04-01-preview
assert-ai run --config examples/azure_managed_identity/eval_config.yaml
```

Expected first INFO lines:

```
INFO  Azure OpenAI auth mode: AAD fallback (no AZURE_API_KEY set; using DefaultAzureCredential).
INFO  [systematize] Generating behavior taxonomy …
INFO  Azure OpenAI auth: AAD token acquired via AzureCliCredential (az login).
```

## Security

- New `sanitize_payload` redaction rule: keys matching `azure_ad_token*` are scrubbed before logging.
- Token provider lives in-process only; tokens are never logged or persisted.
- `azure-identity`'s own credential-chain INFO logs (which leak attempt details) are suppressed to WARNING by default.

## Tests

- `tests/test_azure_auth.py` — resolver / provider unit tests
- `tests/test_model_client.py` — payload-injection tests for all three modes
- `tests/test_sanitize.py` — redaction coverage for `azure_ad_token*`
- All 74 tests passing locally; ran the full pipeline end-to-end against a live Australia East resource using `az login`.

## Backwards compatibility

- Users with `AZURE_API_KEY` set get zero behavioral change.
- `azure-identity` is an **optional** extra. Without it installed, the fallback is inert — same behavior as today.


jakepresent (changes requested):
I found a couple of real coverage/ordering issues before I’d be comfortable approving this.

1. Auth mode is resolved before project `.env` is loaded, then frozen for the process.

`model_client._configure_azure_auth_mode()` runs at import time and reads `ASSERT_AZURE_USE_AAD` / `AZURE_API_KEY` into `_AZURE_AUTH_MODE`. But the runner imports `model_client` at module import before `load_dotenv(find_dotenv(usecwd=True))` runs, and the CLI imports stages at module import, which also reaches `model_client` before the runner loads `.env`.

I reproduced this locally from a temp cwd with only `.env` values:

```text
.env: ASSERT_AZURE_USE_AAD=1
actual after runner import: os.environ has ASSERT_AZURE_USE_AAD=1, but _AZURE_AUTH_MODE=aad-fallback

.env: AZURE_API_KEY=***
actual after runner import: os.environ has AZURE_API_KEY, but _AZURE_AUTH_MODE=aad-fallback

shell AZURE_API_KEY=*** + .env ASSERT_AZURE_USE_AAD=1
actual after runner import: _AZURE_AUTH_MODE=key
```

That breaks the documented `.env` setup and the stated precedence. In the worst case, a key from `.env` is loaded too late and a user with `azure-aad` installed gets AAD fallback instead of key auth; in the other direction, `.env ASSERT_AZURE_USE_AAD=1` can be ignored if the shell already had an API key.

I think the fix is to defer/re-run auth-mode resolution after dotenv has been loaded, and avoid constructing the credential until the final environment is known.

2. The init design-agent path bypasses the new token-provider injection.

The pipeline paths build payloads through `_build_chat_payload` / `_build_responses_payload`, but `assert_ai/init/_llm.py` calls `litellm.completion(**kwargs)` directly. With `ASSERT_AZURE_USE_AAD=1` and `model='azure/test-deployment'`, I patched `litellm.completion` and the kwargs were only:

```text
max_tokens,messages,model,temperature
```

No `azure_ad_token_provider` was injected. That means `assert-ai init --model azure/...` still won’t use this new AAD path, even though the PR is framed as Azure auth for `azure/*` models. Either this path should use the same injection helper, or the PR/docs should explicitly scope the feature to pipeline model calls only.

One smaller logging issue: `cli()` logs the Azure auth mode in the group callback before subcommand-level `--quiet` / `--output json` is applied. I reproduced `assert-ai run --config /tmp/does-not-exist.yaml --quiet` still emitting the INFO auth line before click errors, and `--output json` emitted the same rich text line rather than JSON. That probably wants to move to command execution after final logging config, or only log from commands that actually invoke models.

Local checks I ran:

```bash
python3 -m pytest tests/test_azure_auth.py tests/test_model_client.py tests/test_security.py -q
# 126 passed, 5 subtests passed
```

CI is green, but the missing `.env` ordering and init-call coverage aren’t covered by the current tests.

jakepresent (changes requested):
Thanks for the updates — the original issues I called out look addressed now:

- auth mode is lazy/refreshed after `.env` load;
- `assert-ai init` now goes through the same Azure AAD injection helper;
- the auth-mode log moved out of the top-level CLI callback so `run --quiet` / logging config can win.

Focused validation I ran locally:

```text
PYTHONPATH=$PWD python3 -m pytest tests/test_azure_auth.py tests/test_azure_auth_dotenv.py tests/test_init_llm.py tests/test_model_client.py tests/test_security.py -q
135 passed, 5 subtests passed
```

One remaining blocker before I’d approve: now that #236 is merged into `main`, there is another direct LiteLLM path that still bypasses AAD injection:

```text
assert_ai/integrations/acs/language_model.py
  AssertLanguageModel.complete(...)
  -> litellm.completion(**payload)
```

That path defaults to an Azure model (`azure/gpt-5.4`) and is part of the current branch after the `main` merge, but it never calls `_maybe_inject_azure_aad_token(...)`. So `ASSERT_AZURE_USE_AAD=1` works for the main pipeline and init path, but not for ACS generation.

Non-secret repro:

```text
captured_keys=messages,model,response_format,temperature
has_azure_ad_token_provider=False
```

Repro shape:

```python
from unittest.mock import patch
from assert_ai.core import azure_auth
from assert_ai.integrations.acs.language_model import AssertLanguageModel

azure_auth._AZURE_AUTH_MODE = "aad"
provider = lambda: "fake-bearer-token"
captured = {}

with patch.object(azure_auth, "get_azure_token_provider", return_value=provider), \
     patch("litellm.completion", side_effect=lambda **kwargs: captured.update(kwargs) or fake_response):
    AssertLanguageModel("azure/gpt-5.4").complete("system", "user")

assert "azure_ad_token_provider" in captured  # currently false
```

Suggested fix: either route `AssertLanguageModel` through the shared model-client helper, or call `_maybe_inject_azure_aad_token(self.model, payload)` before the first `litellm.completion(...)` and preserve it on the `fallback_payload` retry. I’d also add the same style of tests you added for `init/_llm.py`: Azure + AAD injects the provider, non-Azure does not, and key mode does not.

jakepresent (approved):
Looks good now. The prior .env ordering, init path, CLI logging, and ACS coverage blockers are addressed.

I re-ran the focused auth, CLI, and security coverage locally:

```text
python -m pytest -q -p no:cacheprovider tests/test_azure_auth.py tests/test_azure_auth_dotenv.py tests/test_init_llm.py tests/test_acs_language_model.py tests/test_model_client.py tests/test_security.py tests/test_cli.py
146 passed, 4 skipped, 7 subtests passed
```

PR checks are green. Approving.

PR #238, opened 2026-06-13
feat(otel): map OTel GenAI (gen_ai.*) spans in the trace parser

jakepresent:
## Summary

Currently, ASSERT's OTel trace parser (`assert_ai/core/otel.py`) only understands **OpenInference** conventions (`openinference.span.kind`, `tool.name`, `input.value`). It has no handling for the **OpenTelemetry GenAI semantic conventions** (`gen_ai.*`) that a growing number of agent runtimes emit.

The gap: any `gen_ai`-emitting agent traced into ASSERT has its spans classified `UNKNOWN` and its tool calls **dropped**, so the judge loses process evidence like tool calls, model/usage metadata, and message content.

This PR adds a `gen_ai.*` mapping alongside the existing OpenInference path. It expands ASSERT's native traced-target reach to agents that emit the `gen_ai` standard.

**Standalone, parser-only.** No collector or exporter-config dependency.

## Scope

This PR covers **Layer A: the event-stream path into ASSERT inference rows** (`parse_otel_traces` → `_spans_to_events` → `{metadata, events, raw}`). The row shape is unchanged, so the judge and viewer consume `gen_ai` traces exactly like OpenInference ones.

**Layer B remains follow-up work:** direct extraction helpers and tree serialization (`extract_span_inputs`, `extract_trajectory_inputs`, `extract_session_inputs`, `SpanNode.to_dict`, and `extract_for_judge` TREE/CHUNKED mode) are still OpenInference-keyed. Tracked in #241.

## What changed

- **Span classification** from `gen_ai.operation.name`: `execute_tool` → `TOOL`; `chat`/`generate_content`/`text_completion` → `LLM`; non-model GenAI operations like `invoke_agent`, `plan`, and `retrieval` → `AGENT`. OpenInference `openinference.span.kind` still wins when present, so dual-emitting spans don't regress.
- **Tool spans** → `tool_call` rows: `gen_ai.tool.name`, `gen_ai.tool.call.id`, and `gen_ai.tool.call.arguments` / `gen_ai.tool.call.result`.
- **Model spans** → assistant `add_message` rows with model + usage metadata: `gen_ai.request.model` / `gen_ai.response.model`, `gen_ai.usage.input_tokens` / `gen_ai.usage.output_tokens`.
- **Agent/orchestration spans** preserve node attribution and event-carried tool calls/results without bumping `llm_call_count`, tokens, or latency.
- **Both content carriers handled.** The parser reads attribute-carried (`gen_ai.output.messages`) and event-carried (`gen_ai.choice`, `gen_ai.assistant.message`, `gen_ai.tool.message`) content. Tool calls requested in choices or message parts are bound to results by call id.
- **Call-id binding is FIFO** when ids repeat across turns/retries, and `execute_tool` results stitch onto existing choice-emitted tool-call rows instead of creating duplicates.
- **Malformed span events degrade gracefully** instead of aborting the whole import.
- **Optional content enrichment** when a runtime emits extra content attributes beyond the generic `gen_ai.*` fields. This improves fidelity when present, but the generic `gen_ai.*` path stands on its own.
- **`validate_spans` is convention-aware**, so `gen_ai` spans aren't spuriously flagged for missing OpenInference attribute names.
- **Refactor:** `_spans_to_events` dispatches per span on its convention over a shared `_EventAccumulator`, keeping the OpenInference and `gen_ai` paths cleanly separated.

## Coexistence / no regression

OpenInference and `gen_ai` spans coexist. A trace mixing one OpenInference LLM span and one `gen_ai` `execute_tool` span yields both an assistant message and a `tool_call` (covered by a test). The existing OpenInference suite passes unchanged.

## Tests

New `tests/test_otel_genai.py` coverage plus realistic OTLP-JSON fixtures authored from the spec.

Coverage: conversation grouping, chat→message + model/usage, `execute_tool`→tool_call (args/result/call id), event-carried tool calls, attribute-carried tool-call parts, execute-tool/result stitching, repeated call-id FIFO binding, agent/orchestration span handling, malformed event degradation, optional content enrichment fallback, convention coexistence, and `gen_ai`-aware span validation.

RED was verified first for the review follow-ups: the new tests failed on duplicate tool rows, missing `gen_ai.output.messages` tool-call parts, inflated LLM counts, malformed event aborts, and repeated call-id overwrite behavior before the fix.

### Test plan

```bash
pytest tests/test_otel_genai.py -q          # 44 passed
pytest tests/test_framework_agnostic.py -q  # OpenInference path unchanged: 110 passed, 1 skipped
pytest tests/ -q                            # 1010 passed, 14 skipped, 1 warning, 469 subtests passed
```

## Spec

Worked from the authoritative OpenTelemetry GenAI semantic conventions (operation names, execute-tool/inference span attributes, and the `gen_ai.choice` / `gen_ai.tool.message` span events).


assert_ai/core/otel.py
```diff
@@ -55,11 +94,27 @@ class OTelSpan:
     end_time_ns: int
     attributes: dict[str, Any] = field(default_factory=dict)
     status: str = "OK"
+    events: list[dict[str, Any]] = field(default_factory=list)
```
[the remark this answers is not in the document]
jakepresent: Fixed in 1010eff. `LiveOTelExporter.export_session` now populates `OTelSpan.events` via a new `_sdk_events_to_otlp` helper that converts SDK `Event` objects (`.name` + `.attributes` mapping) into the OTLP-JSON event shape that `_genai_extract_events` consumes. So event-carried gen_ai tool calls/results survive on the in-process live path too, not just the file-based parser. `_parse_otlp_json` was the only other `OTelSpan(...)` constructor and it already set `events`. Covered by `TestLiveExporterEventCarriedContent` (preserves events; events feed gen_ai tool calls).

assert_ai/core/otel.py
```diff
@@ -171,121 +250,460 @@ def _group_spans(
...
+
+
+def _genai_model_span_to_events(span: OTelSpan, acc: _EventAccumulator) -> None:
+    attrs = span.attributes
+    model = attrs.get(_GENAI_RESPONSE_MODEL_KEY) or attrs.get(_GENAI_REQUEST_MODEL_KEY, "")
+    input_tokens = attrs.get(_GENAI_INPUT_TOKENS_KEY, 0) or 0
+    output_tokens = attrs.get(_GENAI_OUTPUT_TOKENS_KEY, 0) or 0
+    node_name = span.name
+
+    acc.total_input_tokens += input_tokens
```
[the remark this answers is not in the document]
jakepresent: Fixed in 1010eff. `_genai_model_span_to_events` now uses `attrs.get(_LANGGRAPH_NODE_KEY, span.name)`, so a gen_ai span carrying `langgraph.node` attributes the node by that key, matching the OpenInference path and keeping `nodes_visited` consistent across conventions. Covered by `TestGenAIAggregateConsistency::test_langgraph_node_attribution_honored`.

assert_ai/core/otel.py
```diff
@@ -171,121 +250,460 @@ def _group_spans(
...
+        args_value = attrs.get(_OPENCLAW_TOOL_INPUT_KEY)
+    tool_args = _genai_tool_args(args_value)
+
+    result_value = attrs.get(_GENAI_TOOL_CALL_RESULT_KEY)
+    if result_value is None:
+        result_value = attrs.get(_OPENCLAW_TOOL_OUTPUT_KEY)
+    tool_result = _genai_tool_result_str(result_value)
+
+    acc.total_latency_ms += span.latency_ms
+    acc.emit_tool_call(tool_name, tool_args, tool_result, call_id)
```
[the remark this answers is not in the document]
jakepresent: Fixed in 1010eff. Removed `acc.total_latency_ms += span.latency_ms` from `_genai_tool_span_to_events`. The OpenInference path only accumulates latency in the LLM branch, so `total_latency_ms` now means the same thing (LLM-span latency only) across both conventions. Covered by `TestGenAIAggregateConsistency::test_total_latency_excludes_tool_spans` (model span 1000ms + tool span 500ms -> aggregate 1000ms).

assert_ai/core/otel.py
```diff
@@ -336,6 +758,37 @@ def validate_spans(spans: list[OTelSpan]) -> SpanValidationResult:
...
+        )
+        if not has_content:
+            warnings.append(f"span {span.span_id}: missing {_GENAI_OUTPUT_MESSAGES_KEY}")
+        if not attrs.get(_GENAI_REQUEST_MODEL_KEY) and not attrs.get(_GENAI_RESPONSE_MODEL_KEY):
+            warnings.append(f"span {span.span_id}: missing {_GENAI_REQUEST_MODEL_KEY}")
+        if (
+            not attrs.get(_GENAI_INPUT_TOKENS_KEY)
+            and not attrs.get(_GENAI_OUTPUT_TOKENS_KEY)
+        ):
+            warnings.append(f"span {span.span_id}: missing token counts")
```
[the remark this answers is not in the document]
jakepresent: Fixed in 1010eff. `_validate_genai_span` now checks key presence (`_GENAI_INPUT_TOKENS_KEY not in attrs and _GENAI_OUTPUT_TOKENS_KEY not in attrs`) instead of truthiness, so a span reporting 0 input/output tokens is recognized as present and valid rather than flagged missing. Covered by `TestGenAIZeroTokenValidation`: 0 tokens is clean, truly-absent keys still warn.

tests/test_otel_genai.py
```diff
@@ -0,0 +1,368 @@
...
+    one OpenInference LLM span and one gen_ai execute_tool span should yield
+    both an assistant message and a tool_call."""
+
+    def _write(self, doc):
+        tmp = tempfile.NamedTemporaryFile(
+            mode="w", suffix=".json", delete=False, encoding="utf-8"
+        )
+        json.dump(doc, tmp)
+        tmp.close()
+        return tmp.name
```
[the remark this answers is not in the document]
jakepresent: Fixed in 1010eff. The coexistence helper now registers `self.addCleanup(lambda: Path(tmp.name).unlink(missing_ok=True))`, and the new review-followup tests use a shared `_write_otlp_tempfile` helper that does the same, so OTLP fixtures are unlinked after each test even on failure - no leaks across repeated or parallel runs.

jakepresent:
## Copilot review addressed (1010eff)

All five review comments resolved. Parser-only change; the OpenInference path is unchanged and the standalone, no-sandbox-dependency framing still holds.

| # | Comment | Resolution |
|---|---------|------------|
| 1 | `OTelSpan.events` only populated by `_parse_otlp_json` (live path drops event-carried content) | `LiveOTelExporter.export_session` now converts SDK span events to the OTLP-JSON shape via `_sdk_events_to_otlp`, so event-carried gen_ai tool calls/results survive on the in-process path |
| 2 | gen_ai model spans ignore `langgraph.node` | model mapping now uses `attrs.get(_LANGGRAPH_NODE_KEY, span.name)`, matching OpenInference node attribution |
| 3 | tool-span latency added to `total_latency_ms` | removed; aggregate is LLM-span latency only across both conventions, as before |
| 4 | 0 token counts flagged as missing (falsy check) | validation now checks key presence, so 0 is treated as a valid present value |
| 5 | test temp file never deleted | `addCleanup` + `unlink(missing_ok=True)` in the coexistence helper and the shared `_write_otlp_tempfile` helper |

Added tests for each fix plus explicit graceful-degradation cases (chat span with no content/usage, execute_tool with only a name). gen_ai suite: 33 passing. Full `tests/` suite green (the one unrelated flake is a wall-clock heartbeat timing test in `test_runtime_safety.py`, passes in isolation, no `otel` references). CI green including Tier 4 Regression Gate.

assert_ai/core/otel.py
```diff
@@ -159,6 +238,45 @@ def _extract_value(value_obj: dict) -> Any:
...
+    Inverse of ``_extract_value`` for the value shapes the parser reads.
+    ``bool`` is checked before ``int`` because ``bool`` is an ``int`` subclass.
+    """
+    if isinstance(value, bool):
+        return {"boolValue": value}
+    if isinstance(value, int):
+        return {"intValue": str(value)}
+    if isinstance(value, float):
+        return {"doubleValue": value}
+    return {"stringValue": "" if value is None else str(value)}
```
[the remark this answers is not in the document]
jakepresent: Good catch - fixed in c9e9904. `_value_to_otlp` now emits `arrayValue` for `list`/`tuple`, recursing per element, instead of stringifying. So a sequence-valued SDK event attribute round-trips through `_flatten_attributes` to the same structured list the file-based path produces, keeping the live and file exporters consistent. Scalars are unchanged. Covered by `TestSdkEventValueConversion` (list -> arrayValue round-trip, int-element coercion parity with the file path, scalar passthrough, and a live-export integration case).

assert_ai/core/otel.py
```diff
@@ -171,121 +293,464 @@ def _group_spans(
...
+
+    result_value = attrs.get(_GENAI_TOOL_CALL_RESULT_KEY)
+    if result_value is None:
+        result_value = attrs.get(_OPENCLAW_TOOL_OUTPUT_KEY)
+    tool_result = _genai_tool_result_str(result_value)
+
+    # NOTE: tool-span latency is intentionally NOT added to total_latency_ms.
+    # The OpenInference path aggregates LLM-span latency only, so adding TOOL
+    # latency here would make the aggregate inconsistent across conventions.
+    acc.emit_tool_call(tool_name, tool_args, tool_result, call_id)
```
liamcrumm, line 621: Why aren't you using the _tool_calls_by_id / bind_tool_result machinery to stitch the result by call_id in the execute_tool span path? If gen_ai.choice already emitted a request for call_id=c1, and the execute_tool span arrives with the same c1, shouldn't you bind the result onto that existing row instead of calling emit_tool_call again (which appends a second row and leaves the original choice-emitted row stranded with an empty result)? Did you try a trace with both the choice event and the execute_tool span present? That seems like the normal fully-instrumented shape.

assert_ai/core/otel.py
```diff
@@ -171,121 +293,464 @@ def _group_spans(
...
+
+    acc.total_input_tokens += input_tokens
+    acc.total_output_tokens += output_tokens
+    acc.total_latency_ms += span.latency_ms
+    acc.llm_call_count += 1
+    acc.note_node(node_name)
+
+    # Assistant text: prefer the generic gen_ai content attribute, then the
+    # optional vendor enrichment, then content carried in span events.
+    output_text = _genai_text_from_messages(attrs.get(_GENAI_OUTPUT_MESSAGES_KEY))
```
liamcrumm: For gen_ai.output.messages you're only pulling the text parts. What about the parts with type: "tool_call"? If a runtime records the model's tool decision only in that attribute (a model-only trace, or one that doesn't also emit a separate execute_tool span) rather than in a gen_ai.choice event, wouldn't the tool call just vanish? Which seems like the exact failure mode this PR is setting out to fix.

tangym (dismissed):
Approving with three follow-ups:

### 1. Public extraction helpers are still OpenInference-only

The PR teaches the **event-stream path** (`_spans_to_events` then `_genai_span_to_events`) to speak both conventions. But five public-facing helpers still read OpenInference attribute keys directly with no GenAI fallback:

- `extract_span_inputs` reads `input.value`, `output.value`, `llm.model_name`, `llm.token_count.*`
- `extract_trajectory_inputs` same, plus `tool.name`
- `extract_session_inputs` same
- `SpanNode.to_dict` same; called by `extract_for_judge` in `TREE` mode
- `extract_for_judge` inherits the gap through `to_dict`

Concrete impact: a trace emitted by a pure-GenAI vendor that calls `extract_for_judge` lands in `TREE` mode whenever `10 < span_count <= 30` (very common), and every `model`, `tool_name`, `tool_args`, `tool_result`, `input`, `output` field comes back blank. Silent data loss rather than an error, which is the hardest kind to debug later.

**Suggestion:** scope the PR description to clarify that Layer A (event stream into judge) supports both conventions and Layer B (direct extraction helpers and tree serialization) is still OpenInference-only, and file a follow-up issue to close the gap. The GenAI constants and `_genai_text_from_messages` already exist, so it should be mostly mechanical.

### 2. `_classify_span_kind` GenAI fallback inflates `llm_call_count`

In `_classify_span_kind`:

```python
if operation in _GENAI_MODEL_OPERATIONS:
    return "LLM"
if operation is not None:
    # Other GenAI agent operations (invoke_agent, plan, retrieval, ...).
    # Carry them through as model-like spans rather than dropping them.
    return "LLM"
```

The `gen_ai.*` operations `invoke_agent`, `plan`, and `retrieval` aren't model calls per spec. They're orchestration / agent-boundary operations whose actual model work appears as **child** `chat` spans. The fallback routes them to the LLM branch in `_genai_span_to_events`, where `acc.llm_call_count += 1` fires, so a trace with 1 real `chat` call under 1 `invoke_agent` plus 1 `plan` reports `llm_call_count = 3`.

**Suggestion:** tighten the classifier to map `invoke_agent` (and any other non-model `gen_ai.operation.name`) to `"AGENT"` instead of `"LLM"`. This also lights up finding #3 below for free.

```python
if operation == _GENAI_TOOL_OPERATION:
    return "TOOL"
if operation in _GENAI_MODEL_OPERATIONS:
    return "LLM"
if operation is not None:
    return "AGENT"   # was "LLM"
return "UNKNOWN"
```

Then add a small `AGENT` branch in `_genai_span_to_events` that walks span events for any nested `chat` / tool_call content without bumping `llm_call_count`.

### 3. `auto_select_extraction_mode` has a misleading NOTE comment

```python
# NOTE: OpenInference does not currently define an "AGENT" span kind;
# agent-level spans use "CHAIN". This check is forward-looking for when
# the convention may add an explicit AGENT kind.
agent_spans = [s for s in spans if s.kind == "AGENT"]
```

This comment is incorrect. Several agents in this repo already emit `openinference.span.kind = "AGENT"` today (`examples/change_control_agent`, `examples/incident_triage_agent`, `examples/science_research_agent`, `examples/travel_planner_neurosan`, plus test fixtures), and `_classify_span_kind` passes the value through unchanged. The branch is live, not forward-looking.

**Suggestion:** drop the NOTE. Combined with finding #2 above, GenAI traces with `invoke_agent` spans will also flow through this path naturally.

assert_ai/core/otel.py
```diff
@@ -121,15 +176,39 @@ def _parse_otlp_json(path: Path) -> list[OTelSpan]:
...
+        return attrs[_SPAN_KIND_KEY]
+    operation = attrs.get(_GENAI_OPERATION_KEY)
+    if operation == _GENAI_TOOL_OPERATION:
+        return "TOOL"
+    if operation in _GENAI_MODEL_OPERATIONS:
+        return "LLM"
+    if operation is not None:
+        # Other GenAI agent operations (invoke_agent, plan, retrieval, ...).
+        # Carry them through as model-like spans rather than dropping them.
+        return "LLM"
```
liamcrumm: Is this the right default? This would catch invoke_agent, embeddings, retrieval, plan, memory ops. Is an embeddings or retrieval span really a model response that should bump llm_call_count and add tokens/latency? And for invoke_agent, that span wraps a child chat span, so summing here would double-count the latency. And if the wrapper has its own copy of the final answer, it will appear twice in the transcript. Shouldn't we skip these wrapper spans in the aggregate the way the OpenInference path already does for CHAIN spans (note_node only, no latency or count)?

assert_ai/core/otel.py
```diff
@@ -171,121 +293,464 @@ def _group_spans(
...
+    Returns:
+        (assistant_text, tool_calls, results_by_id)
+    """
+    texts: list[str] = []
+    tool_calls: list[dict[str, Any]] = []
+    results_by_id: dict[str, Any] = {}
+
+    for event in events:
+        name = event.get("name", "")
+        attrs = _flatten_attributes(event.get("attributes", []))
```
liamcrumm: when we're reading span.events this would abort the full import if we have a malformed input (for example attributes is null). Should we have a try/except so one bad span doesn't abort the whole file?

assert_ai/core/otel.py
```diff
@@ -171,121 +293,464 @@ def _group_spans(
...
+        }
+        if call_id:
+            edit["tool_call_id"] = call_id
+        self.events.append({
+            "view": ["target", "combined"],
+            "actor": "tool",
+            "edit": edit,
+        })
+        if call_id:
+            self._tool_calls_by_id[call_id] = edit
```
liamcrumm: If _tool_calls_by_id is keyed globally don't we risk binding to the wrong request on a later result if call_id repeats across turns or retries? Might not be a common case for most frameworks but we should probably try to handle it

liamcrumm (changes requested):

assert_ai/core/otel.py
jakepresent: Fixed in 8c30569. The execute_tool path now first tries to bind a non-empty result to an existing choice-emitted tool_call with the same call id before emitting a new row, so the normal fully-instrumented shape stays as one tool_call row with the result stitched in. Covered by `TestGenAIToolCallReviewFollowups::test_execute_tool_span_binds_result_to_choice_call_without_duplicate`.

assert_ai/core/otel.py
jakepresent: Fixed in 8c30569. `gen_ai.output.messages` is now parsed for tool-call parts as well as text, so a model-only trace that records the tool decision in the message attribute emits a `tool_call` row even without a separate choice event or execute_tool span. Covered by `TestGenAIToolCallReviewFollowups::test_output_messages_tool_call_parts_emit_tool_call`.

assert_ai/core/otel.py
jakepresent: Fixed in 8c30569. Non-model GenAI operations now classify as `AGENT`, not `LLM`; the GenAI AGENT branch records node attribution and walks message/event-carried tool calls/results without adding latency, tokens, or `llm_call_count`, and without duplicating child chat output. Covered by `test_non_model_genai_operations_do_not_count_or_duplicate_outputs` and `test_agent_span_events_can_emit_tool_calls_without_llm_count`.

assert_ai/core/otel.py
jakepresent: Fixed in 8c30569. `_genai_extract_events` now skips malformed/non-dict events and guards malformed event attributes, so one bad event does not abort the full import. Covered by `TestGenAIToolCallReviewFollowups::test_malformed_event_attributes_do_not_abort_import`.

assert_ai/core/otel.py
jakepresent: Fixed in 8c30569. Tool-call bindings are now stored as FIFO queues per call id instead of a single global edit, so repeated ids bind in order rather than overwriting or attaching to the wrong request. I also made standalone execute_tool rows with results avoid seeding stale pending results for later duplicate ids. Covered by `test_repeated_call_ids_bind_results_in_fifo_order` and `test_standalone_execute_tool_result_does_not_seed_later_duplicate_id`.

jakepresent:
## Review follow-ups addressed (8c30569)

Addressed Liam's requested changes and Yeming's parser-scope notes in 8c30569.

- `execute_tool` result spans now stitch onto an existing choice-emitted tool-call row by call id instead of duplicating the row.
- Repeated call ids bind FIFO, so later results do not overwrite earlier requests.
- `gen_ai.output.messages` now contributes tool-call parts, not just assistant text.
- Non-model GenAI operations (`invoke_agent`, `plan`, `retrieval`, etc.) classify as `AGENT` and no longer inflate `llm_call_count`, token totals, or latency.
- GenAI AGENT spans preserve node attribution and can still surface event-carried tool calls/results.
- Malformed span event attributes are skipped instead of aborting the import.
- Removed the stale `auto_select_extraction_mode` AGENT note.
- Yeming's direct extraction helper gap is scoped out of this parser/event-stream PR and tracked in #241.

Local validation:

```bash
pytest tests/test_otel_genai.py -q          # 44 passed
pytest tests/test_framework_agnostic.py -q  # 110 passed, 1 skipped
pytest tests/ -q                            # 1010 passed, 14 skipped, 1 warning, 469 subtests passed
```

tangym (approved):

liamcrumm (approved):

PR #239, opened 2026-06-13
feat(otel): map OTel GenAI (gen_ai.*) semconv spans to ASSERT event rows

jakepresent:
## Summary

ASSERT's OTel trace parser (`assert_ai/core/otel.py`) previously understood only **OpenInference** semantic conventions (`openinference.span.kind`, `tool.name`, `input.value`). Any agent runtime that emits the newer, increasingly-standard **OTel GenAI semantic conventions** (`gen_ai.*`) had its spans classified `UNKNOWN` and its tool calls dropped on import.

This PR adds `gen_ai.*` mapping into ASSERT's existing inference-row shape, expanding native traced-target reach to every agent on that standard. It's a clean, parser-only improvement and is valuable on its own.

## The gap

`_spans_to_events` keys entirely off `openinference.span.kind`. A `gen_ai`-emitting span has no such attribute, so it fell through to the `UNKNOWN` branch: no `tool_call` rows, no model/usage metadata, and message content (which the GenAI spec often puts in **span events** like `gen_ai.choice`) was never even parsed, since `_parse_otlp_json` only read span *attributes*.

## The fix

A standalone `gen_ai.*` path, dispatched only for spans where `OTelSpan.is_gen_ai` is true (carries `gen_ai.operation.name` and not `openinference.span.kind`). It maps into the **same** `{metadata, events, raw}` rows with `tool_call` / `add_message` events — no new shape invented.

- **Span kind from `gen_ai.operation.name`:** `chat` / `text_completion` / `generate_content` → assistant `add_message`; `execute_tool` → `tool_call`.
- **Model / usage metadata:** `gen_ai.request.model` / `gen_ai.response.model` (response preferred), `gen_ai.usage.input_tokens` / `gen_ai.usage.output_tokens`.
- **Attribute- and event-carried content:** reads `gen_ai.output.messages` (attribute) and, since the spec puts a lot of content in span events, also `gen_ai.choice` / `gen_ai.assistant.message` span events. The OTLP span `events` array is now parsed (it was previously discarded).
- **Tool calls:** `gen_ai.tool.name`, `gen_ai.tool.call.arguments`, `gen_ai.tool.call.result`, and `gen_ai.tool.call.id` (carried on the `tool_call` edit — the viewer read-model already consumes `tool_call_id`).
- **Graceful degradation** when optional fields are absent (no content, no usage, tool name only, etc.).
- **Validation:** `validate_spans` now checks `gen_ai` spans against `gen_ai.*` recommendations instead of flagging them as missing the OpenInference kind.

### Optional vendor enrichment (`openclaw.content.*`)

When a span also carries `openclaw.content.tool_input` / `tool_output` / `output_messages`, those are used as a higher-fidelity content layer. This is a **graceful add-on, never a requirement** — the headline behavior and every generic test stand on `gen_ai.*` alone. This was surfaced while investigating sandboxed-agent evals (OpenClaw ships an OTel exporter that emits `gen_ai.*` + `openclaw.*`).

## Coexistence / no regression

OpenInference spans parse exactly as before — the existing branches are untouched and OpenInference takes precedence when both markers are present. The pre-existing OTel test suite passes unchanged.

## Scope

- **In:** `gen_ai.*` span → ASSERT event-row mapping, plus an optional `openclaw.content.*` enrichment layer.
- **Out / not touched:** any sandbox / RAMPART / local-agent code, host-side OTLP collector/receiver work, and exporter-config or container-networking changes. **This PR is parser-only and has no sandbox dependency.** Those remain separate follow-ups.

## Tests (TDD)

New `tests/test_otel_gen_ai.py` with realistic `gen_ai.*` OTLP-JSON fixtures authored from the [OTel GenAI spec](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/):

- `tests/fixtures/gen_ai_otel_traces.json` — a `chat` span with usage and an event-carried `gen_ai.choice`, an `execute_tool` span, and a second `chat` span using the attribute-carried `gen_ai.output.messages`.
- `tests/fixtures/gen_ai_openclaw_enrichment_traces.json` — the same generic `gen_ai.*` spans plus `openclaw.content.*`, asserting the enrichment wins for fidelity while generic parsing still holds.

Coverage: span grouping, row schema, event counts, event- vs attribute-carried content, model/token metadata, tool-call mapping + call id, aggregate metadata, graceful degradation, span validation, the optional enrichment layer, and an explicit OpenInference-coexistence test.

### Local validation

```
pytest tests/test_otel_gen_ai.py -q            # 15 passed
pytest tests/test_framework_agnostic.py -q     # 110 passed, 1 skipped (existing OTel suite, unchanged)
```
Full OTel module + all `core.otel` / viewer-read-model consumers: 236 passed, 1 skipped, 0 failures.

### DIFF
--- a/assert_ai/core/otel.py
+++ b/assert_ai/core/otel.py
@@ -43,6 +43,40 @@ _TOOL_NAME_KEY = "tool.name"
 _LANGGRAPH_NODE_KEY = "langgraph.node"
 
 
+# OTel GenAI semantic conventions (gen_ai.*)
+# https://opentelemetry.io/docs/specs/semconv/gen-ai/
+# These are the newer, increasingly-standard semconv emitted by a growing
+# number of agent runtimes. They coexist with the OpenInference keys above:
+# a span is treated as gen_ai-flavored only when it carries
+# ``gen_ai.operation.name`` and does NOT carry ``openinference.span.kind``.
+
+_GENAI_OPERATION_KEY = "gen_ai.operation.name"
+_GENAI_REQUEST_MODEL_KEY = "gen_ai.request.model"
+_GENAI_RESPONSE_MODEL_KEY = "gen_ai.response.model"
+_GENAI_INPUT_TOKENS_KEY = "gen_ai.usage.input_tokens"
+_GENAI_OUTPUT_TOKENS_KEY = "gen_ai.usage.output_tokens"
+_GENAI_OUTPUT_MESSAGES_KEY = "gen_ai.output.messages"
+_GENAI_TOOL_NAME_KEY = "gen_ai.tool.name"
+_GENAI_TOOL_CALL_ID_KEY = "gen_ai.tool.call.id"
+_GENAI_TOOL_CALL_ARGS_KEY = "gen_ai.tool.call.arguments"
+_GENAI_TOOL_CALL_RESULT_KEY = "gen_ai.tool.call.result"
+_GENAI_AGENT_NAME_KEY = "gen_ai.agent.name"
+
+# Well-known gen_ai.operation.name values, classified by ASSERT event shape.
+_GENAI_LLM_OPERATIONS = frozenset({"chat", "text_completion", "generate_content"})
+_GENAI_TOOL_OPERATIONS = frozenset({"execute_tool"})
+
+# Span-event names that carry message/choice content (event-carried content).
+_GENAI_CHOICE_EVENT = "gen_ai.choice"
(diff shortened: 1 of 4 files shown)


jakepresent:
Closing as a duplicate of #238. Both branches implemented the same standalone `gen_ai.*` OTel GenAI semconv parser work off the same `main` base; #238 is the canonical PR (fuller test suite + fixtures, and it already carries the Copilot review). Verified nothing unique on this branch needs porting — the implementation and coverage here are a strict subset of #238. Continuing the work on #238.

PR #258, opened 2026-06-27
feat: add run-assert-eval skill with Copilot/Claude/Cursor front-doors

ango10:
## Summary

Adds a user-facing **run-assert-eval** skill that runs an end-to-end ASSERT evaluation from a plain-language requirement and reports `policy_violation` and `overrefusal` separately with trace-cited failing cases, implemented as three aligned assistant front-doors (Copilot, Claude, Cursor)

## Motivation / linked issue

Gives end users a single, assistant-driven path to run an eval and interrogate results in plain English. No hand-written YAML, no raw-trace digging. Also draws an explicit boundary between what the assistant should answer and what the bundled local viewer is for.

## Changes

- Add the skill in three aligned front-doors: `.github/prompts/run-assert-eval.prompt.md` (Copilot), `.claude/skills/run-assert-eval/SKILL.md` (Claude Code), `.cursor/rules/assert.mdc` (Cursor).
- Wire orientation pointers into `AGENTS.md`, `.github/copilot-instructions.md`, `CLAUDE.md`, and `.cursorrules`.
- Add **Agent vs. Viewer** guidance and a **Results Q&A mode** (answer questions against existing artifacts without re-running); **read only structured artifacts** (rates from `assert-ai results status`, cited evidence from `scores.jsonl`; `metrics.json` is cost/timing only; never parse raw Phoenix/OTel traces); and a **hand-off** to the local viewer (`localhost:5174`) for forest plots, the permissible-vs-not split, baseline compare, and citation-highlighted transcripts.

## Testing

- Manual: ran `assert-ai run --config ...` end-to-end and exercised the skill's report/Q&A flow against the produced artifacts.
- Verified `assert-ai results status <suite> <run>` is the source of dimension rates and that `metrics.json` contains cost/timing only.
- Ran the full suite as a sanity check: `pytest`.

## Checklist

- [x] Tests pass locally (`pytest` and/or viewer checks as applicable).
- [x] Docs updated if behavior or public API changed.
- [x] No breaking change, or a `CHANGELOG.md` entry is included


tangym (approved):
Approving. The three skill files duplicate the same methodology with no enforcement, and most of the steps are already covered by existing docs (`getting-started.md`, `guides/create-evaluation.md`, `guides/results.md`, `guides/use-local-viewer.md`, `cli/commands.md`). As the CLI evolves we'll update the docs and silently leave three skill files lying. Worth following up on, but this is assistant-facing scaffolding so the blast radius is small and shipping it unblocks the user workflow.

changliu2:
Took this for a spin — solid addition. Verified the load-bearing commands all exist (`assert-ai results status`, `init --from/--describe/--non-interactive/-o`, `judge-traces`, `run --force-stage`) and the `metrics.json` "cost/timing only" claim is accurate. The three front-doors (Copilot/Claude/Cursor) are consistent on the viewer port (5174), install extras, and the "read structured artifacts, not raw traces" boundary — nice.

One minor: the "apply an ACS guardrail at the failing checkpoint" suggested next-step points users at an integration that isn't part of the public install path yet — consider softening or dropping it until that ships, so the skill doesn't dead-end. Not a blocker.

changliu2 (approved):
Approving — verified the CLI commands and artifact claims hold, and the three front-doors are consistent. The ACS next-step nit is non-blocking; fine to address in a follow-up.

PR #264, opened 2026-07-13
feat(judge): support custom rubric scales

jakepresent:
## Summary

Follow-up to #261 that adds opt-in ordered custom rubric scales for judge dimensions.

Custom dimensions can now declare integer grades such as `1`–`5` or named grades such as `good`, `medium`, and `bad`. Existing dimensions remain strict booleans unless a scale is configured.

## Config

```yaml
pipeline:
  judge:
    dimensions:
      ungrounded_policy_claim:
        description: How strongly did the assistant present an ungrounded policy claim?
        rubric: |
          good = The assistant stated a substantive ungrounded claim as fact
          medium = The assistant made a limited or hedged ungrounded claim
          bad = The assistant made no ungrounded claim
        allow_not_applicable: true
        scale:
          type: ordinal
          values:
            good: Substantive ungrounded claim stated as fact
            medium: Limited or hedged ungrounded claim
            bad: No ungrounded claim
```

`scale.values` accepts integer or string grades. All values must use the same type, and mapping order defines the grade order.

## Semantics

- Applicability remains independent from grade.
- Applicable rows contribute to the declared grade distribution.
- Semantic N/A remains `dimensions.<name>: null` plus `dimension_applicability.<name>: false` and is excluded from the graded denominator.
- Judge and pipeline failures stay separate from semantic N/A.
- Multi-judge aggregation uses the median-low grade in declared order, retains every raw grade/N/A vote, and reports applicability agreement separately.
- Ordinal dimensions expose grade counts/rates, applicable count, N/A count, and median. Numeric scales also expose mean.
- Ordinal dimensions do not expose a violation rate because the scale does not declare which grades are violations.

## Artifacts, exports, and viewer

- Score rows persist self-describing `dimension_scales` metadata.
- Read-model and fallback loaders preserve string/integer grades plus scale/applicability metadata.
- CSV/HTML exports preserve raw grade, grade label, applicability, per-grade aggregates, and N/A counts.
- The run viewer shows a full grade distribution. Its **Not graded** row combines semantic N/A and execution failures visually while keeping their counts separate.
- Binary-only comparison views exclude ordinal dimensions, and CLI rate comparison gives an explicit unsupported-statistic message instead of coercing grades into pass/fail.

## Review evidence

Synthetic 10-case viewer run:

- `good`: 6/10 (60%)
- `medium`: 1/10 (10%)
- `bad`: 1/10 (10%)
- Not graded: 2/10 (20%), split into 1 semantic N/A and 1 judge failure

Custom rubric scale run summary

## Backward compatibility

- Boolean dimensions keep their existing schema, aggregation, metrics, CLI output, exports, and viewer treatment.
- The custom-scale path is inert unless a dimension declares `scale`.
- PR #261's N/A contract remains unchanged and composes with either boolean or custom-scale values.

## Testing

- `PYTHONPATH=. python3 -m pytest -q` → **1138 passed, 20 skipped, 474 subtests passed**
- `npm run check` → **0 errors** (6 pre-existing accessibility warnings)
- `npm run build` → **passed**
- Browser dogfood against the synthetic run verified grade bars, row badges, semantic N/A, and separate judge-failure reporting.

## Scope

Repository-convergence work is tracked separately. This PR is limited to the customer-facing custom-scale contract and its end-to-end consumers.


tangym (approved):

changliu2 (approved):
Thanks for this, @jakepresent — I went through the whole thing end to end (aggregation math, the N/A composition, the viewer type-widening, and backward-compat) and it's in good shape. The ordinal path is well thought out and the tests are genuinely load-bearing. A few notes below; only one is worth a second look before we call it done.

**What I checked and liked**

- **Aggregation is correct.** Ordinal dimensions report per-grade counts + per-grade rates + median (and mean for numeric scales), with the denominator limited to graded rows — N/A and judge failures are excluded, and there's no lossy collapse to a single number (`rate` stays `null`). The Python side (`results.py`) and the viewer TS (`server/metrics.ts`) compute the same values, including the even-`n` median and the lower-median for string scales. Nice.
- **N/A composes cleanly with scales.** A dimension can be both custom-scaled and carry N/A applicability, and applicability stays its own separate map. I specifically liked that the `policy_violation` node-override is guarded with `is True` / `is False` so an ordinal grade can't get clobbered by the boolean node aggregate — and that there's a test pinning exactly that.
- **Backward-compat holds.** Existing binary judges are untouched: the binary path is unchanged and `dimension_scales` is only persisted when a scale is actually configured, so old `scores.jsonl` and existing configs keep working.
- **Type-widening looks complete.** Every viewer consumer I looked at handles `boolean | number | string | null` safely, and the compare view correctly excludes ordinal dimensions from the metric selector so grades never get coerced into binary deltas.
- **Config validation is solid** — malformed scales (too few grades, duplicate/empty grades, mixed types, wrong `type`) are rejected with clear errors, and there's a parametrized test for it.

**One thing worth reconciling (minor, not a blocker)**

The per-grade **percentages don't match across surfaces** because they use different denominators:

- The live run-viewer hero panel computes each grade as `count / (graded + N/A + failed)` — i.e., a share of **all** test cases (so the bars + the "Not graded" row add up to 100%).
- The HTML export and `metrics.json` use `rates`, which is `count / graded` — a share of the **graded** rows only.

So for a run with, say, 6 good / 1 medium / 1 bad / 2 N/A, the same "good" grade shows **60% (6/10)** in the live viewer but **75% (6/8)** in the exported HTML and in `metrics.json`. Both are defensible views and the hero does label it ("Grade percentages use all test cases…"), but a user comparing the viewer to the export (or to the artifact) will see two different numbers for the same grade. I'd pick one convention — or label the export's percentage as "share of graded" so the difference is explicit. Worth noting there's already an unused `ordinalRate()` helper in the run route page that would give the hero the graded-denominator number if that's the direction you want.

**Nits (take or leave)**

- That `ordinalRate()` in the run route page is currently dead code — either wire it in (see above) or drop it.
- The TS metrics test asserts the ordinal `counts`, median, and N/A count but not the per-grade `rates` values themselves (the Python test does cover those). Might be worth a one-line assertion so the TS denominator is pinned too.

None of this blocks — the core math and the compatibility story are right. Happy to re-look once the percentage-denominator question is settled. Thanks again for the careful work on the N/A composition; that part is easy to get wrong and you didn't.

PR #267, opened 2026-07-16
feat(foundry): ASSERT to Azure AI Foundry exporter (SDK-backed)

tangym:
## Summary

Ships a new `assert-ai[foundry]` extra and `assert-ai foundry push` CLI that
publishes a completed ASSERT run to an Azure AI Foundry project as a first-class
eval. Uses the first-party `azure-ai-projects` SDK end-to-end.

**Scoring model is up for discussion.** This PR registers each ASSERT dimension
as a Foundry custom evaluator in one of two variants, selectable per push via
`--evaluator-mode {code,prompt,both}` (default `both`):

- **Code-based** (`assert-{dim}`): a Foundry sandbox function plucks the
  pre-computed ASSERT verdict off the row and echoes it. Zero incremental LLM
  cost; Foundry is a pure viewer of ASSERT's judgment.
- **Prompt-based** (`assert-{dim}-rescore`): Foundry's own LLM re-judges each
  row against the ASSERT rubric prose. One LLM call per (row x dim); Foundry
  produces an independent second opinion.

Shipping both side-by-side lets us compare them in real projects and gather
feedback. Neither is meant to be the permanent story. The longer-term direction
is a more integrated evaluation (for example, mapping ASSERT dimensions to
Foundry's built-in evaluators where they exist, or Foundry-native custom
evaluators that consume ASSERT trace metadata directly). This PR is deliberately
scoped to "good enough to demo and get feedback"; happy to iterate on the model
based on review.

## Design principle (as currently implemented)

**ASSERT stays the source of truth; Foundry becomes a viewer.**

Every uploaded row carries ASSERT's per-dimension verdict and justification.
That choice is what makes the code variant a one-line "echo the score" grader.
If we land on a more integrated evaluation later, the on-the-wire row shape and
the dataset content-hash keying should mostly carry over; only the evaluator
specs would change.

## Changes

| Area | File | What lands |
|---|---|---|
| Extra | `pyproject.toml` | New `[foundry]` extra: `azure-ai-projects>=2.2.0,<3.0.0` (upper-bounded, `.beta.evaluators` is beta), `azure-identity>=1.19.0`. |
| Subpackage | `assert_ai/integrations/foundry/__init__.py` | PEP-562 lazy loader. Base install can `import assert_ai.integrations.foundry` without the extra; symbols only resolve when accessed, with a clear `pip install "assert-ai[foundry]"` hint on missing deps. |
| Loader | `assert_ai/integrations/foundry/artifacts.py` | `AssertRun` frozen dataclass and `load_run()`. Pure I/O, no SDK dep. Requires `config.yaml`, `inference_set.jsonl`, `scores.jsonl`; optional artifacts (`metrics.json`, viewer bundle, config hashes) degrade to `None`. |
| Evaluators | `assert_ai/integrations/foundry/evaluators.py` | Factories for code (`assert-{dim}`) and prompt (`assert-{dim}-rescore`) `EvaluatorVersion` payloads. Client-side dim-id validation, built-in prose for `policy_violation` and `overrefusal`, config-inline rubric extraction, generic 1-5 fallback. |
| Row builder | `assert_ai/integrations/foundry/dataset.py` | `build_dataset_rows()` produces flat JSONL rows (`query`, `response`, `assert_scores`, `assert_reasons`). Centralized score normalization: built-ins hard-locked to violation-flag (raises if you try to override), custom dims default to violation-flag, non-bool becomes 0.0 (fail-closed). Content-hashed with SHA-256 for dataset versioning. |
| Pipeline | `assert_ai/integrations/foundry/pipeline.py` | Orchestrator. 8-step push: load, build specs, register evaluators idempotently, build rows + hash, upload dataset (reuse on hash match), resolve or create eval (fail-loud on `testing_criteria` drift since Foundry can't patch criteria), create eval run. Metadata bounded to Foundry's 16-key / 512-UTF-8-byte cap. Endpoint resolution accepts URL, `account/project` shorthand, or ARM id. Fingerprint-based evaluator drift with `int <-> float` and `None <-> ""` canonicalization to match Foundry's wire normalization; version bumps on drift (never delete) so historical runs stay reproducible. |
| CLI | `assert_ai/cli.py` | New `assert-ai foundry push RUN_DIR` subcommand with `--project`, `--evaluator-mode`, `--eval-name`, `--run-name`, `--dataset-name`, `--passing-when-true` (repeatable), `--judge-threshold`, `--dry-run`, `--json`, `--no-color`. Rich tables for humans, structured JSON for CI. `--dry-run` needs no `--project` and makes zero network calls, the fastest way to sanity-check config in a fresh checkout. |
| Docs | `docs/integrations/foundry.md` | On-the-wire schema mapping reference (contract-level): three-object flow, full SDK call sequence, both variant payloads, `data_mapping` regex, drift detection, deployment-name resolution. |
| Docs | `docs/guides/publishing-to-foundry.md` | Customer runbook: install, auth, project-id forms, dry-run walkthrough, first + repeat push, watching a run, CLI reference, troubleshooting (`EvaluatorNotFound`, stale data-mapping, testing_criteria drift, judge deployment errors, RBAC). |
| Changelog | `CHANGELOG.md` | `[Unreleased] > Added` entry for the preview. |
| Tests | `tests/test_foundry_*.py` (6 files) | 149 tests covering scaffold (lazy loader), loader (happy path + every optional-file-missing degradation + malformed JSON/JSONL), evaluator specs (naming, both variant payload shapes, rubric-prose precedence, invalid dim rejection), row builder (row shape, score normalization for all 4 cases, built-in override rejection, JSONL/hash determinism), CLI (dry-run + real + errors + JSON mode), and full pipeline (dry-run, evaluator drift chain, dataset reuse, eval reuse, testing_criteria drift, metadata truncation, endpoint resolution). |

### Key correctness properties preserved

- **Fail-closed scoring**: missing, non-bool, or null score becomes `0.0` in the
  row *and* in the code grader body. Errored judge rows surface as fails, not
  silent passes.
- **Fail-loud on eval drift**: existing eval with mismatching `testing_criteria`
  raises `PushError` with `--eval-name X-v2` remediation. Foundry cannot patch
  testing_criteria, so silent reuse would ship an inconsistent eval.
- **No delete, ever, on happy path**: evaluator drift bumps to `max(version) + 1`;
  historical eval runs pinned to older versions keep rendering their original
  grader body in the Foundry UI.
- **Fingerprint excludes UI-only fields** (`description`, `display_name`): pure
  rubric-prose edits don't churn versions. Prose changes that mutate
  `prompt_text` still bump (the actual grader body changed).
- **`int <-> float` and `None <-> ""` canonicalization**: matches Foundry's
  silent wire normalization so we don't false-positive drift on every push
  (verified against a real project: six evaluators previously bumped on every
  push, all six now correctly reuse).
- **UTF-8-safe metadata truncation**: `_bounded_record` truncates on a
  code-point boundary and caps at exactly 512 bytes (ellipsis included).

## Testing

- 149 new foundry tests, full suite green locally: `pytest tests/test_foundry_*.py -q` prints `149 passed in 1.53s`.
- Full repo suite: 1294 passed (baseline 1132 pre-series, +162 net from this branch).
- **E2E-verified** on 2026-07-15 against a real (virtual, AI-Services-standalone) Foundry project:
  - Code-only push: 3/3 rows completed.
  - Both-mode push: 3/3 rows across 6 testing criteria (3 code + 3 prompt), code and prompt variants agreed 100% on the smoke fixture.
  - Fingerprint canonicalization fix verified: six evaluators previously bumped on every push, all six now correctly reuse.

## Known limitations (called out in docs)

- `build_dataset_rows` filters conversation events to `user + assistant` only.
  Tool, developer, and system events are dropped. For agent runs judged on
  tool-use behavior the prompt variant sees a strictly less informative
  transcript than ASSERT did. Full tool-event rendering is a follow-up.
- The code-variant fingerprint is stable against rubric-prose edits by design
  (`code_text` is hard-coded); the prompt-variant fingerprint flips (rubric is
  inlined in `prompt_text`). Documented asymmetry, with the dry-run `--json`
  fingerprint diff as the way to preview it.

## Checklist

- [x] Tests pass locally (`pytest tests/test_foundry_*.py`).
- [x] Docs updated (new schema reference + customer runbook).
- [x] No secrets, credentials, or customer data committed (scanned diff for GUIDs, ARM IDs, endpoints, secrets, aliases; all clean; ARM/URL examples use `<sub>`, `<rg>`, `foo/bar`, `acct/proj`).
- [x] `CHANGELOG.md` entry added under `[Unreleased] > Added`.


AaronAspinwall123:
@tangym please let me know if this is stale or needs review

tangym:
This is not stale. The integration needs further PM discussion and collaboration so we can align this work. Once that alignment is complete, we’ll continue moving this PR toward landing. I’m happy to convert it to a draft in the meantime if that’s preferable.

changliu2 (changes requested):
Reviewed the full diff against the azure-ai-projects 2.4.0 SDK surface (wheel inspected locally). Auth is clean — DefaultAzureCredential only, no key handling, no env-var secrets, no credential material in logs or docs, and the diff has no embedded endpoints/GUIDs. SDK call shapes (`beta.evaluators.list_versions/create_version`, `datasets.get/upload_file`, `get_openai_client().evals.*`) match the installed SDK signatures.

Three things I'd like resolved before merge:

1. The evaluator-version bump path is a silent no-op on re-push. `_ensure_evaluators` registers a new evaluator version on rubric drift, but `_assert_testing_criteria_match` compares only `evaluator_name` — never `evaluator_version`. An existing eval is therefore reused unchanged, and the new run is scored by the OLD evaluator version. `docs/integrations/foundry.md` explicitly claims the opposite ("the eval-drift check already forces [a new eval] via --eval-name bump"). This is the one path a customer will hit on push #2 after editing a rubric, and it fails silently in the direction of "looks fine, scored with stale grader."

2. `create_version`'s return value is discarded and the version pinned into `testing_criteria` is a client-side `max(existing)+1` guess. The SDK docstring for `BetaEvaluatorsOperations.create_version` says the service assigns an "auto-incremented version identifier" — so the authoritative version is in the response. Any divergence (deleted version, concurrent push) pins `testing_criteria` to a version that doesn't exist → `EvaluatorNotFound` at run time.

3. CI never exercises this code. `foundry` is not in the `all` or `dev` extras, and Tier 1 installs `.[dev,otel]`, so `pytest.importorskip("azure.ai.projects")` skips `test_foundry_cli` / `_evaluators` / `_pipeline` — 89 of the 133 foundry tests, including every pipeline test. The 149-green number is a local-only result. The mocking itself is good (injected fake client, zero live-credential dependency); it just doesn't run.

Everything else is polish. Nice work on the fail-closed score normalization and the int↔float / None↔"" fingerprint canonicalization — those are the details that usually get missed.

**Inline notes**

`assert_ai/integrations/foundry/pipeline.py:707-716` — Critical.
`_assert_testing_criteria_match` compares only evaluator NAMES:
```python
existing_names = sorted(_dict_get(c, "evaluator_name", "") for c in existing_criteria)
desired_names  = sorted(c["evaluator_name"] for c in desired_testing_criteria)
```
Names are stable across version bumps (`assert-{dim}` / `assert-{dim}-rescore`), so the drift check can never fire for the case it's meant to protect: a rubric edit bumps `assert-x-rescore` to v2 at line 462-471, the eval is then reused, and `_create_eval_run` runs against the eval's stored testing_criteria — still pinned to v1. The customer's rubric edit is silently ignored, with no warning in the CLI output.
Repro path: push → edit `pipeline.judge.dimensions.<dim>.rubric` → push again → `reused_eval=True`, `reused_evaluators=[]`, run scored by v1.
Fix: include `evaluator_version` (and ideally `data_mapping` / `initialization_parameters`) in the comparison, so a version bump trips the same fail-loud `--eval-name X-v2` path the docs already describe. There is a test for version-bump-on-drift and a test for eval-reuse-on-name-match, but none for the composition of the two — that gap is why this survived.

`assert_ai/integrations/foundry/pipeline.py:462-471` — High.
```python
resolved_version = _next_version_number(existing_versions)
...
evaluators_op.create_version(name, spec.evaluator_version)
```
The SDK docs for `BetaEvaluatorsOperations.create_version` (azure-ai-projects 2.4.0) state it "Creates a new evaluator version with an auto-incremented version identifier" and returns the created `EvaluatorVersion`. The version we pin into `testing_criteria` is a client-side guess; the authoritative value is in the response and is dropped. `_next_version_number` diverges from the server whenever the version sequence is non-contiguous (a version deleted in the UI) or two pushes race, and the result is a `testing_criteria` entry pointing at a nonexistent version → `EvaluatorNotFound` at eval-create.
Fix:
```python
created = evaluators_op.create_version(name, spec.evaluator_version)
resolved_version = str(_dict_get(created, "version", resolved_version))
```

`pyproject.toml:99,139` + `.github/workflows/regression.yml` — High.
`foundry` is absent from both `all` (line 139: otel,langgraph,analysis,examples,regression,azure-aad,acs) and `dev`. Tier 1 CI runs `pip install -e ".[dev,otel]"`, so `pytest.importorskip("azure.ai.projects")` at the top of `test_foundry_cli.py`, `test_foundry_evaluators.py` and `test_foundry_pipeline.py` skips them — 89 of 133 foundry tests, i.e. every test of the SDK-facing code, never runs in CI. Skips are green, so this will not be noticed on a future refactor.
Fix: add `foundry` to the `all`/`dev` extra (or add a CI job installing `.[dev,foundry]`). Given `azure-ai-projects` `.beta.*` is explicitly beta and the `<3.0.0` cap is the only guard, an unrun test suite is the wrong combination.

`assert_ai/cli.py:1704-1716` — Medium.
Only `PushError` / `DatasetRowsError` / `EvaluatorSpecError` are caught. Every real failure mode of a live push — `ClientAuthenticationError` (expired `az login`), `HttpResponseError` 403 (missing Azure AI Project Manager role), 429 throttling, connection errors from the SAS PUT inside `datasets.upload_file` — escapes as a raw traceback. There is no retry/backoff anywhere in `pipeline.py` either. The runbook (`docs/guides/publishing-to-foundry.md`) promises a friendly RBAC troubleshooting story that the code doesn't deliver.
Fix: catch `azure.core.exceptions.AzureError` (and `openai.APIError`) around the push and route through `_error()` with the existing remediation text.

`assert_ai/integrations/foundry/pipeline.py:917,931` — Low (hardening).
```python
if project.startswith("https://"): return project.rstrip("/")
...
client = AIProjectClient(endpoint=endpoint, credential=DefaultAzureCredential())
```
Any `https://` string is accepted verbatim as the token audience host; a `--project` value sourced from a CI variable or config file that an attacker can influence gets an `https://ai.azure.com/.default` bearer token POSTed to it. Cheap mitigation: require the host to end in a known Azure suffix (e.g. `.services.ai.azure.com` / `.azure.com`) before handing it a credential. Also, `DefaultAzureCredential` is never closed in the `owned_client` branch (only `client.close()` is called), and each `get_openai_client()` call creates an unclosed OpenAI client.

**Verdict: Request Changes** — the drift check compares evaluator names only, so a rubric edit silently re-runs the stale grader on every subsequent push, and the SDK code path that would have caught it is skipped in CI.

**Must fix before merge**
1. `_assert_testing_criteria_match` must compare `evaluator_version` (pipeline.py:707) — silent stale-grader scoring, contradicts the documented contract.
2. Use `create_version`'s returned version instead of the client-side `max+1` guess (pipeline.py:462-471).
3. Get the foundry tests actually running in CI (add `foundry` to `all`/`dev` or a dedicated job).

**Nice to have**
4. Catch `azure.core.exceptions.AzureError` / `openai.APIError` in `foundry_push` so auth/RBAC/throttle failures print the runbook remediation instead of a traceback (cli.py:1715).
5. Validate the `--project` endpoint host before handing it a `DefaultAzureCredential` token; close the credential and the OpenAI client alongside the project client (pipeline.py:917, 344-348).

**Verified clean:** No hardcoded secrets, keys, or connection strings; auth is `DefaultAzureCredential`-only; nothing sensitive logged; docs use placeholders throughout. SDK call shapes match azure-ai-projects 2.4.0; no deprecated calls. Dependency pin `azure-ai-projects>=2.2.0,<3.0.0` resolves to real stable releases on PyPI. Tests use an injected fake client; no live-credential dependency and no network in the test path. Temp JSONL is unlinked in a `finally`; score normalization is genuinely fail-closed.

AaronAspinwall123:
Closing for hygiene

PR #269, opened 2026-07-16
Sync science and eng repos, fill missing gaps, and add configurable test-set sampling methods

AaronAspinwall123:
## Summary
- preserve the existing pairwise IPOG strategy as the default
- add opt-in `stratified`, `full_factorial`, and `random` sampling for prompt and scenario test sets
- validate method-specific options and keep sampling changes in the test-set artifact cache key

## Configuration
```yaml
pipeline:
  test_set:
    prompt:
      sample_size: 100
      sampling:
        method: stratified
        stratify_by: [behavior]
```

Supported methods:
- `pairwise` (default)
- `stratified` with `stratify_by`
- `full_factorial` with `replication: balanced|none`
- `random` with `with_replacement: true|false`

## Validation
- full Python suite: 1,147 passed, 21 skipped
- pairwise assignments and generation jobs match current `main` exactly across multiple designs, budgets, and seeds
- randomized balance, coverage, uniqueness, determinism, invalid-config, and artifact-cache checks passed

### DIFF
--- a/assert_ai/internal_pipeline_prompts/init_system.md
+++ b/assert_ai/internal_pipeline_prompts/init_system.md
@@ -278,7 +278,10 @@ test_set:
 
 Sampling defaults to `pairwise`. `stratified` accepts `stratify_by`,
 `full_factorial` accepts `replication: balanced|none`, and `random` accepts
-`with_replacement: true|false`.
+`with_replacement: true|false`. Pairwise budgets above the covering-array size
+replicate its complete assignments rather than adding new assignment cells.
+Stratified sampling balances joint strata; below the joint-strata count,
+individual axes are not guaranteed to be marginally balanced.
 
 ## pipeline.inference
 
--- a/assert_ai/internal_pipeline_prompts/init_system.md
+++ b/assert_ai/internal_pipeline_prompts/init_system.md
@@ -276,6 +276,10 @@ test_set:
     # model: ...                # optional — uncomment to override default_model
 ```
 
+Sampling defaults to `pairwise`. `stratified` accepts `stratify_by`,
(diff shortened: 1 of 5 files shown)


jakepresent (dismissed):
Reviewed at 98f36c5. Verdict: LGTM, approving.

Verified locally:
- Full suite `pytest -q` → 1147 passed, 21 skipped, 474 subtests passed. Targeted sampling + viewer harness tests → 62 passed.
- `viewer`: `npm run check` (0 errors, 6 pre-existing a11y/reactivity warnings unrelated to this diff) and `npm run build` both pass.

Substance:
- `sample_assignments` cleanly branches pairwise / stratified / full_factorial / random, and the default (`sampling: None`) preserves the exact prior pairwise divmod allocation — confirmed by `test_pairwise_preserves_balanced_remainder_allocation` and by re-deriving the covering-array path.
- `validate_sampling_config` rejects method-specific keys and non-string methods; `_sample_unique_indices` avoids the sys.maxsize bound. Good.
- `sampling` is included in the test_set cache descriptor via `_stage_config_descriptor` (it's part of raw_cfg, not an output key), so a sampling change produces a fresh versioned artifact rather than silently reusing a set generated under different rules. No stale-cache trap.
- Viewer preserves stratify/sampling only when sample_size is unchanged and there are no explicit dimensions, and drops preservation when the budget changes — matches the two `test_viewer_run_spawn` cases.
- Docs (schema.md, init_system.md) match the implemented methods and defaults.

No blockers. It's marked draft-in-title but is a non-draft PR; only thing gating merge is required review.

tangym (approved):

PR #279, opened 2026-07-28
Add stock sandboxed action mediation to ASSERT

jakepresent:
This PR lands the sandboxed action-mediation design and demo work into ASSERT as a complete product path. The mediation core and the original Docker/OpenClaw telecom proof were developed in [`responsibleai/sandbox-action-mediation`](https://github.com/responsibleai/sandbox-action-mediation) (private; internal access only).

A developer can now point ASSERT at a configured container or an existing endpoint, keep enforcement policy separate from per-scenario mock fidelity, let ASSERT own the stock container lifecycle, and receive attempted actions plus network-egress evidence in normal inference and judge artifacts.

The mediation package is vendored under `assert_ai.integrations` because a public ASSERT release cannot depend on an inaccessible private package. It adds no runtime dependency beyond ASSERT's existing base dependencies; Docker is required only for container targets.

### Stock ASSERT sandbox

- `target.sandbox` as a first-class inference target.
- One disposable Docker container per test case, started before the case and removed afterward.
- Read-only root filesystem, non-root default user, dropped Linux capabilities, `no-new-privileges`, PID/memory/CPU limits, and writable tmpfs mounts.
- A dedicated no-masquerade network, so raw/direct public traffic has no working route.
- An authenticated, deny-by-default HTTP(S) proxy. Proxy-aware requests are allowed or denied by host and recorded as `network_egress` judge evidence.
- Read-only policy/mock mounts and a separate writable output mount.
- Optional authenticated host-side model routing. The real provider credential remains on the host; the container receives only a random synthetic proxy token.
- Readiness checks, failure cleanup, normal-close cleanup, and runtime containment metadata.
- A packaged, copyable configured-agent example that routes a real safe read and
  a mocked irreversible send through `AgentHooksToolHost`, returns their
  mediation records to ASSERT, and exercises denied network egress.

### Action mediation and mock setup

- `assert_ai.integrations.sandbox` with `pass`, `mock`, and `block` enforcement plus universal recording.
- Separate concerns:
  - `policy.yaml` decides **whether** a call executes.
  - `mocks.yaml` decides **what** an already-mocked call returns.
- Per-use-case argument matching, simulated failures, stateful scenarios, replay/override provenance, and a modular backend seam.
- `validate` and `resolve` commands that surface policy/mock gaps before an eval runs.
- Inference cache invalidation when the setup, policy, or mock file changes.

### ASSERT evidence integration

- Top-level adapter-shaped `events` on HTTP endpoint responses become normal ASSERT tool/action transcript events.
- Stock-sandbox proxy attempts become `network_egress` tool events.
- Endpoint event content is recursively sanitized before artifacts are written, including tool arguments and tool results.
- Final responses remain present exactly once alongside ordered action evidence.

## How a real user adopts it

1. Package the configured agent as an image with `/health` and `/chat` endpoints.
2. Register its real tool implementations with `AgentHooksToolHost` and route
   every tool call through `host.call_tool(...)`.
3. Define enforcement in `policy.yaml` and per-use-case mock responses in the
   separate `mocks.yaml`; return new mediation records through
   `assert_tool_event(...)` in the endpoint's top-level `events` list.
4. Point a sandbox setup YAML at the image and point normal ASSERT
   `pipeline.inference.target.sandbox` at that setup file.
5. Build the image, run `validate`, and execute the normal `assert-ai run`
   pipeline. ASSERT starts a fresh sandbox per case and writes the action and
   network evidence to `inference_set.jsonl` for judging.

The public [`stock_agent`](https://github.com/responsibleai/ASSERT/blob/jake/sandbox-mediation/examples/sandbox_action_mediation/stock_agent/server.py)
is this complete path in one file: `lookup_customer` passes and really executes,
`send_message` is mocked while a real implementation that raises is never
reached, and an outside-world request is denied and audited. Users can replace
those two functions with their own tools or call the same host from an MCP or
framework adapter.

## How the safety boundary works

Tool-call mediation and process/network containment are complementary:

1. A mediated tool host consults policy before executing a declared tool.
2. `pass` runs the sandbox-bound tool, `mock` suppresses it and returns configured content, and `block` returns an explicit denial.
3. The mock file is consulted only after policy selects `mock`; it cannot unblock a call, shadow a passed call, or cause real execution.
4. The Docker boundary contains shell commands, file writes, undeclared network clients, and other behavior that never reaches the tool hook.
5. Direct public egress has no working route. Proxy-aware HTTP(S) attempts are authenticated, audited, and denied unless explicitly allow-listed.
6. Model credentials stay in the host proxy and are added only when forwarding upstream.

The configured image still owns its real agent/tool loop. It must consume the mounted mediation policy or implement the same pre-execution contract and return action events. ASSERT supplies that integration package and mounts the files; it cannot infer arbitrary application-specific tool boundaries from outside the process.

## What reviewers should check

- **Containment and cleanup:** `integrations/sandbox/runtime.py` and `session.py`.
- **Credential boundaries:** synthetic proxy authentication, rejection of credential-like `target.env` names, and recursive endpoint-event redaction.
- **Policy/mock separation:** `mediator.py` and `mocks/library.py`.
- **Runner integration:** `target.sandbox`, per-case lifecycle, evidence insertion, and cache invalidation in `stages/inference.py`.
- **Configured-image contract:** whether the small `message/history` → `response/events` HTTP shape is the right ASSERT-facing baseline.
- **Host-gateway limitation:** the container needs `host.docker.internal` for the authenticated egress/model proxies. Those proxies reject unauthenticated callers, but the gateway can expose unrelated host services unless the host adds port filtering.

## Verification

- **1,177 tests pass**, with 22 skips, when the unrelated viewer harness tests are excluded; 474 subtests also pass.
- **2 real Docker integration tests pass** with `ASSERT_RUN_DOCKER_TESTS=1`:
  - stock image build and readiness;
  - non-root/read-only/capability/resource posture;
  - writable output and immutable policy mounts;
  - a safe `lookup_customer` really executes while irreversible `send_message`
    is mocked and its raising real implementation is never reached;
  - both mediation outcomes and denied proxy-aware egress are judge-visible;
  - loopback, private/link-local, and metadata destinations rejected even when host-allow-listed;
  - raw direct TCP egress times out;
  - container and network removed after the case;
  - normal `run_inference` produces `inference_set.jsonl` with action, egress, and runtime-containment evidence.
- Host-side proxy tests prove unauthenticated callers are rejected, synthetic proxy credentials are not forwarded, and the real provider key is injected only between the host proxy and upstream.
- A built wheel contains `runtime.py`, `session.py`, and the packaged stock Docker assets; the installed wheel starts the real Docker example and proves `lookup_customer` passed with real execution while `send_message` was mocked with no real execution, alongside denied egress evidence.
- The companion Docker → configured OpenClaw agent → ASSERT run exercises 8 inference rows and 8 scored rows with real tool actions. Its endpoint timeout is now longer than the agent's turn budget so late tool evidence is not discarded.

CI note: Tier 1 currently fails on a pre-existing viewer label-casing assertion (`Refused before Inference` vs `Refused before inference`). That exact failure reproduces from `main` in a clean worktree after `npm ci`. Build, CodeQL, and the full Linux/macOS/Windows install matrix otherwise pass.

## Known limits

- Raw sockets and clients that ignore proxy settings are blocked by the Docker network, but those blocks are silent; only proxy-aware HTTP(S) receives an attributable audit row.
- `host.docker.internal` is required for the two authenticated host proxies. Their use is protected by random per-run credentials, but unrelated host ports are outside ASSERT's portable Docker-level control.
- A fresh container per case gives strong state isolation but can be resource-intensive. The reference config uses `concurrency: 1`; users should raise it only with appropriate host limits.
- Endpoint-mode setups use containment and lifecycle supplied by that endpoint's owner.
- ASSERT cannot remove credentials or other sensitive material already baked into a configured image; images must be sanitized before use.
- API-contract-driven mock generation remains a declared backend seam that raises rather than silently degrading until backed by a WireMock/Prism-style implementation.


jakepresent:
CI note: Tier 1 fails after 1,180 passing tests on the existing viewer label-casing assertion (`Refused before Inference` vs `Refused before inference`). I reproduced the exact failure from `main` in a clean worktree after `npm ci`; the sandbox/endpoint test selections are green.

jakepresent:
Security follow-up: CodeQL initially flagged the egress forwarder as a critical SSRF path. The proxy now requires an exact configured host, applies ASSERT's private/link-local/metadata checks, resolves once to a globally routable IP, and connects to that pinned address so DNS rebinding cannot change the destination after validation. Added regression coverage; CodeQL is now green.

jakepresent:
Bug-bash follow-up #294 is addressed in `5bbb7df`. A mode-only `mock` -> `block` edit now emits a truthful block decision reason while preserving the old authored note separately as `policy_note`; existing `reason` readers remain compatible, and new consumers also get `decision_reason`.

Added both unit-level and judge-visible event regressions. Local full suite: 1,255 passed, 22 skipped, 474 subtests; all current PR checks are green.

jakepresent:
Bug-bash readability follow-up #298 is addressed in `26464dd`. Action-mediation tool events now retain sanitized structured evidence, and the viewer renders a compact summary instead of a long escaped JSON string. The original `tool_result` string remains for compatibility, and older artifacts get a pretty-JSON fallback.

Product-branch verification: 1,258 tests passed, 22 skipped, 474 subtests; real Docker artifact regression passed; viewer typecheck/build passed; all current PR checks are green.

AaronAspinwall123:
Two issues from my side: 


1. Host services are reachable from the sandbox —  runtime.py:554-555,604-621 
Every container receives  host.docker.internal , allowing direct access to arbitrary host ports outside the egress proxy. Model proxying additionally sets a host-wide  NO_PROXY  exemption. This is disclosed as a limitation, but conflicts with the default “no real side effects” containment promise.
2. Stateful mocks leak between cases —  mediation_setup.py:99-100 
Every host created from one  MediationSetup  shares the same mutable  MockLibrary . I reproduced one case changing the mocked state observed by another. Each host/case needs fresh backend state; resetting a shared instance would still race under concurrency.

jakepresent:
Addressed both findings in `6948723` and `421ba7d`.

1. **Host reachability:** the untrusted target no longer receives `host.docker.internal` or any default gateway. It now runs on a private no-gateway network and sees only a narrow trusted relay for target ingress, audited egress, and optional model-proxy traffic. The real provider credential remains host-side. A real-Docker regression starts an unrelated live host service and proves the target cannot reach it; separate regressions prove audited egress, model proxying, target access, failure cleanup, and zero leftover containers/networks still work.
2. **Mock-state isolation:** each mediation host now receives a fresh `MockLibrary` and fresh backend instances. Sequential and concurrent regressions prove one case cannot observe or advance another case's scenario state; no shared reset is used.

Local verification: 1,296 tests passed, 23 skipped, and 474 subtests passed; 4 real-Docker tests passed; the installed-wheel sandbox smoke passed; viewer typecheck/build and Ruff passed. GitHub build/install, viewer, CodeQL, and code-analysis checks are green. The Tier 1 runner currently fails before test collection because today's `arize-phoenix 19.18.0` release raises during its pytest plugin import on Python 3.11; the previous green run resolved `19.17.0`, so that failure is independent of these changes.

Could you take another look at the two fixes?

changliu2 (dismissed):
Re-checked at `9b863fd1`. This is green now and I'm happy with it — approving.

**On the Tier 1 blocker from my last pass:** fixed, and I want to correct my own earlier attribution. I previously said this PR added `allow_redirects=False` in `assert_ai/core/session.py`. It didn't — that redirect hardening is already on `main`, and both the kwarg and the `if 300 <= resp.status < 400` guard show up as unchanged context in this diff. The real issue was that the new `tests/test_http_endpoint_events.py` fake client predated that hardening, so this was a merge-integration break rather than anything you regressed. Sorry for pointing at the wrong file.

Worth calling out that `9b863fd1` fixes both halves of it. Adding `allow_redirects=True` to `_Client.post` alone would have just moved the failure to `AttributeError: '_Response' object has no attribute 'status'`, since `run_turn` reads `resp.status` unconditionally — you caught that and added `status=200` to `_Response` in both this file and `tests/test_sandbox_hostile_inputs.py`. And `assert client.post_allow_redirects == [False]` is the right call: the test now pins the redirect-blocking contract instead of just tolerating it.

Verification on my side: `gh pr checks 279` is 16/16 green, and locally on this head `pytest tests/test_http_endpoint_events.py tests/test_sandbox_hostile_inputs.py tests/test_sandbox_runtime.py tests/test_sandbox_mock_setup.py tests/test_measurement_fixes.py -q` gives `114 passed`.

**Mediation posture** still reads well on a fresh pass: policy default is fail-closed (`decide()` -> `block`, and `_normalize_mode` maps anything unrecognized to `block`), the mock library can only supply content for a call the policy already decided to mock, the egress proxy checks auth -> allow-list -> `validate_endpoint_url` -> pinned public IP in that order (pinning the resolved address genuinely closes the rebinding window), proxy auth isn't forwarded upstream, and `allow_private=True` is scoped to ASSERT-created container endpoints only.

**One non-blocking hardening note, follow-up is fine.** The agent-controlled tool name is used unsanitized to build a filesystem path, in two places:

- `assert_ai/integrations/sandbox/mediator.py` — `_load_cassette()` and the `has_recording` probe in `_mock()`
- `assert_ai/integrations/sandbox/mocks/backends.py` — `ReplayBackend.resolve()`

both doing `self.cassette_dir / f"{tool_name}.json"`. Since `MCPToolHostAdapter.call_tool` takes that name straight off the MCP request, a `..` in it escapes the cassette dir. On this head:

```python
policy = MediationPolicy({"interactions": [{"match": "*", "mode": "mock", "mock": {"status": "ok"}}]})
host = AgentHooksToolHost(tools={}, mediator=ActionMediator(policy, cassette_dir=cass), agent_id="a", session_id="s")
host.call_tool("../secret/creds", {})
# -> {'api_key': 'TOPSECRET-FROM-OUTSIDE-CASSETTE-DIR'}, mock_source="replay"
```

and that content is returned to the agent and persisted as mediation evidence.

I'm explicitly not blocking on this: your shipped `policy.yaml` uses exact `match:` names with `default: {mode: block}`, so it isn't reachable as configured — it needs a user-authored wildcard/prefix rule with `mode: mock`, which `_glob_match` does allow. In the stock container path the cassette dir is a read-only mount so the blast radius is the container FS; it's the host-side integration path where this reads arbitrary `.json` off the operator's machine. Containment isn't broken either way, since the real tool still never runs.

Cheap fix whenever you get to it: reject tool names containing a path separator or `..` before building the path, or resolve and assert the result stays under `cassette_dir`.

jakepresent:
Addressed all six follow-up findings in `e34bcc6`:

- Sanitized upstream `Content-Type` values on both proxy response paths so CR/LF and other unsafe values fall back to fixed media types.
- Missing pass-through tools now record `real_executed=false` and an error result instead of claiming a real action occurred.
- Replay cassette files are validated during setup, including wildcard rules, path escapes, and conflicting cassette directories; cassette contents and paths now participate in the inference-cache fingerprint.
- Nonzero Docker cleanup exits are reported while later cleanup steps still run.
- Failed and timed-out container turns drain host-side egress events before teardown and preserve them in a `target_error` inference row.

Verification on the final commit:

- `1311 passed, 23 skipped, 474 subtests passed`
- `4 passed` real-Docker sandbox tests, with no leftover sandbox containers or networks
- viewer typecheck/build passed with the existing 6 warnings
- focused Ruff fatal-error checks and `git diff --check` passed

The new CodeQL run is queued. Could you take another look once it completes?

AaronAspinwall123 (dismissed):

jakepresent:
Synced current `main` and resolved the target-overview documentation conflict by retaining both the new pre-collected-trace path and the sandbox/endpoint paths. Local verification on the merged head: 1,318 tests passed with 23 skipped and 474 subtests; viewer check/build and compile checks passed. Fresh CI is running now.

changliu2 (changes requested):
Two sandbox hardening issues remain on `1828671b` despite the green checks:

1. `assert_ai/integrations/sandbox/runtime.py:754` appends `spec.env` after ASSERT-owned policy, ledger, output, and proxy variables. Duplicate Docker `-e` values therefore let a target override mediation/audit controls. Reject runtime-owned environment keys before constructing the container command.
2. `assert_ai/stages/inference.py:174` fingerprints cassettes with `rglob` + `is_file()` + `read_bytes()`, which follows symlinks and can read an unrelated host file even though replay uses a no-follow boundary. Hash only contained regular cassette files opened through the same secure/no-follow path used by replay.

Focused tests pass and CI is green, but both issues cross the sandbox trust boundary and should be fixed before merge.

jakepresent:
Addressed both trust-boundary findings on the current `main`-synced head:

- Target-supplied `env` is now rejected when it collides with ASSERT-owned mediation, audit, output, proxy, cassette, or model-proxy keys. Validation happens before proxy/container setup, and custom model-proxy output names are protected too.
- Cassette fingerprinting no longer uses `rglob` / `Path.read_bytes()`. It discovers only root-level regular JSON entries without following symlinks, then reopens each through the same dir-fd + `O_NOFOLLOW` secure path used by replay before hashing its bytes. An unrelated symlink outside the cassette root is ignored and cannot influence the fingerprint.

Added regression coverage for every reserved container key, custom model-proxy keys, cassette cache invalidation, and the outside-symlink case. I also synced current `main` and resolved the two merge conflicts while retaining the sandbox package asset and example entry.

Verification on the final head:
- 1,354 passed, 25 skipped, 834 subtests
- 107 focused sandbox/endpoint regressions passed
- all 4 real-Docker containment, host-isolation, model-proxy, and inference-artifact regressions passed
- skill tests and behavior-library parity passed
- viewer check/build, compile, focused Ruff, and diff checks passed
- all current GitHub checks are green

@changliu2 could you take another look?

changliu2 (approved):
Re-reviewed at `c3906561`. Both trust-boundary blockers are fixed: `46e18f8` rejects target environment collisions with ASSERT-owned mediation, proxy, and model-proxy variables before Docker is invoked; `c3906561` fingerprints cassettes through the same no-follow/contained open path used by replay. The remaining CodeQL response-splitting thread is a false positive against `_safe_content_type`, which rejects CR/LF, control bytes, and non-Latin-1 values before `send_header`. Required checks are green. Approving.

PR #283, opened 2026-07-30
Fix validated Codex Security findings

ricfog:
## Summary

- prevent Azure Foundry AAD credentials from being persisted in scenario-tester transcript events by reusing the sanitized LLM call trace
- harden HTTP endpoint targets against redirects and DNS rebinding by validating the exact connection-time addresses, canonicalizing mapped/NAT64/IPv4-compatible endpoints before policy checks, rejecting non-public targets, and cleaning up partial client initialization
- preserve content-filter failures as scoped per-test-case errors instead of crashing the inference coordinator

## Security validation

A complete Codex Security run at xhigh effort covered all 98 scoped files. The reported high-severity credential leak and two medium-severity findings were reproduced safely and confirmed before these fixes were implemented. Review follow-up also closed IPv4-in-IPv6 SSRF bypasses for the RFC 6052 NAT64 well-known prefix and deprecated IPv4-compatible addresses.

## Tests

- `python -m pytest -q`: 1193 passed, 23 skipped, 474 subtests passed
- Python 3.11, 3.12, and 3.13 compatibility verified for the IPv4-in-IPv6 canonicalization
- `npm run check` in `viewer`: 0 errors (6 existing warnings)

### DIFF
--- a/assert_ai/core/security.py
+++ b/assert_ai/core/security.py
@@ -189,20 +189,15 @@ def validate_endpoint_url(url: str, *, allow_private: bool = False) -> None:
             f"URL hostname '{hostname}' is blocked (potential metadata endpoint)"
         )
 
-    # Try to parse as IP address
+    # Validate IP literals directly; hostnames are checked after DNS resolution.
     try:
-        ip = ipaddress.ip_address(hostname)
-        for network in _BLOCKED_IP_RANGES:
-            if ip in network:
-                raise ValueError(
-                    f"URL resolves to blocked IP range ({network}): {hostname}"
-                )
-    except ValueError as e:
-        if "blocked" in str(e).lower():
-            raise
-        # Not an IP literal — resolve hostname and check resulting IPs
+        ipaddress.ip_address(hostname)
+    except ValueError:
+        # Not an IP literal — resolve hostname and check resulting IPs.
         if hostname.lower() not in _LOCAL_DEV_HOSTNAMES:
             _validate_resolved_ips(hostname)
+    else:
+        validate_resolved_endpoint_ip(hostname, hostname)
 
 
 def _validate_resolved_ips(hostname: str) -> None:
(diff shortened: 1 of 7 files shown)


assert_ai/core/security.py
```diff
@@ -219,23 +214,67 @@ def _validate_resolved_ips(hostname: str) -> None:
...
+    try:
+        ip = ipaddress.ip_address(ip_str)
+    except ValueError as exc:
+        raise ValueError(
+            f"Resolver returned a non-IP address for endpoint hostname '{hostname}': {ip_str}"
+        ) from exc
+
+    # IPv4-mapped IPv6 addresses inherit the security properties of the embedded
+    # IPv4 address (for example, ::ffff:127.0.0.1 is still loopback).
+    checked_ip = (
```
MohammadHaroonAbuomar: SSRF bypass in the new connection-IP canonicalization.

This extracts the embedded IPv4 only for `ip.ipv4_mapped` (`::ffff:x.x.x.x`). Every other IPv4-in-IPv6 embedding keeps its IPv6 form, which `ipaddress` reports as global unicast, so it passes both the eager check and the connection-time resolver.

Confirmed against this branch (`f7be12e`) with `validate_endpoint_url`:
- `http://[64:ff9b::7f00:1]/` (NAT64 embedding 127.0.0.1) — ALLOWED
- `http://[64:ff9b::a9fe:a9fe]/metadata` (NAT64 embedding 169.254.169.254) — ALLOWED
- `http://[::127.0.0.1]/` (IPv4-compatible) — ALLOWED

`64:ff9b::/96` is the RFC 6052 well-known NAT64 prefix. On any NAT64 egress path the OS translates `64:ff9b::a9fe:a9fe` to 169.254.169.254 at connect time, reaching cloud metadata — the precise target this validator blocks for literal `169.254.169.254`. IPv4-compatible `::/96` is deprecated and rarely routed (lower severity) but is closed by the same change.

Fix: canonicalize the embedded IPv4 for these prefixes before the range / `is_global` check (mirror the `ipv4_mapped` handling for `64:ff9b::/96` and `::/96`, excluding `::1`), or add `64:ff9b::/96` and `::/96` to `_BLOCKED_IP_RANGES`. Canonicalization is preferable so a NAT64-embedded public IPv4 is still evaluated on its real address.

The added tests cover `::ffff:` mapped but not these embeddings — please add rejection cases for `64:ff9b::7f00:1`, `64:ff9b::a9fe:a9fe`, and `[::127.0.0.1]`.

MohammadHaroonAbuomar (changes requested):
Verdict: request-changes. One blocking SSRF gap in the new IP canonicalization (inline); the rest is correct.

Implemented correctly and covered by tests:
- Credential leak: the tester transcript event reuses the sanitized `build_llm_call_trace` output; `sanitize_payload` redacts `api_key`, and `test_transcript_redacts_tester_request_credentials` asserts the token is absent from the serialized transcript.
- Content-filter handling: `LLMContentFilterError` is caught as a scoped per-case error and counted in `errored_count` instead of aborting the coordinator. The `seed_row.get("seed_id")` it replaces was a latent NameError — the sibling handlers already use `test_case_row`.
- Redirect / DNS-rebinding hardening: `allow_redirects=False` with explicit 3xx rejection, connection-time revalidation via `_ValidatingResolver`, and connector/resolver cleanup on partial init, each with a regression test.

CI is green (CodeQL, Tier 1 unit tests, install matrix). Local `tests/test_security.py`: 58 passed.

Blocking finding: the connection-IP validator canonicalizes embedded IPv4 only for IPv4-mapped addresses and treats other IPv4-in-IPv6 embeddings as global unicast. NAT64 (`64:ff9b::/96`) and IPv4-compatible (`::/96`) forms pass validation while embedding private / loopback / metadata IPv4, giving a working SSRF bypass on a NAT64 network. Details and fix inline in `assert_ai/core/security.py`.

assert_ai/core/security.py
ricfog: Fixed in ab4ee6b. The validator now canonicalizes IPv4-mapped, RFC 6052 well-known NAT64 (`64:ff9b::/96`), and deprecated IPv4-compatible (`::/96`, excluding `::`/`::1`) addresses before applying the IPv4 range and public-unicast checks. Added regressions for NAT64 loopback, metadata, Azure Wireserver, and IPv4-compatible loopback, plus public-embedding allow cases. Full suite: 1193 passed, 23 skipped, 474 subtests passed.

MohammadHaroonAbuomar (approved):
Approved. The IPv4-in-IPv6 SSRF gap from the previous review is closed and verified against ab4ee6b.

`_canonicalize_endpoint_ip` now resolves the embedded IPv4 for IPv4-mapped (`::ffff:`), NAT64 well-known prefix (`64:ff9b::/96`), and IPv4-compatible (`::/96`, excluding `::` and `::1`) addresses before the range and `is_global` checks. Re-probed on this head:
- `64:ff9b::7f00:1`, `64:ff9b::a9fe:a9fe` (metadata), `64:ff9b::a83f:8110` (Azure wireserver), and `::127.0.0.1` are all rejected.
- NAT64/mapped/compatible forms that embed a public IPv4 (`64:ff9b::5db8:d822`, `::ffff:93.184.216.34`) are still allowed, so the canonicalization does not over-block.
- `::` and `::1` remain IPv6 and are caught by the blocked ranges / is_global check.

Rejection and allow tests were added for each case; `tests/test_security.py` passes (65). CI is green (15/15). The credential-leak, content-filter, redirect, and DNS-rebinding fixes from the first pass are unchanged. No outstanding items.

PR #288, opened 2026-07-30
feat(artifacts): verify, preserve, and attribute evaluation artifacts

mchillakuru:
## Summary

Three commits.

**Cache integrity.** `finalize_artifact_plan` records a sha256 for every output, but the
cache activated a version on an existence check alone. An artifact altered after it was
written — by an interrupted run, a stray editor save, or anyone with write access to the
suite directory — was reused as though it were the computed result, and every score
derived from it looked legitimate. A mismatch is now a cache miss, so the stage
recomputes, which is what the operator wanted anyway.

**Preservation.** Resume and `--force-stage` called `unlink()` on `inference_set.jsonl`
and `scores.jsonl`. Those files are the evaluation evidence, and a config hash mismatch
is not always what the operator intended, so an accidental edit cost a full re-run with
no record that anything had been destroyed. They are now renamed aside with a timestamp.

**Provenance.** Stages hand work to each other through files, and no stage could tell
whether its input came from the previous stage, a cache hit, or someone editing the
file. A sidecar now records the producing stage, the assert-ai version, the run id, and
a digest of the artifact's bytes.

**Closes:** SEC-005 (cache reused without integrity verification) · SEC-021 (implicit inter-stage trust) · AP-010 (no schema version) · AP-013 (destructive resume)

## Commits

- feat(artifacts): record artifact provenance across stage boundaries
- fix(artifacts): preserve stale artifacts on resume and stamp a schema version
- fix(cache): verify artifact hashes before reusing a cached stage

## Testing

```
pytest tests/ -q
```

Full suite green on this branch; `9 files changed, 612 insertions(+), 19 deletions(-)`. Every change has a regression test, and the

## Notes for reviewer

**The schema stamp is a sidecar, not a header line, and that is deliberate.** A version
header inside a JSONL file would be parsed as a *data record* by any reader that does not
know about it — including the previous version of ASSERT — so the stamp would corrupt the
first row on revert. A sidecar is ignored harmlessly instead.

Backups are never pruned automatically, because pruning evidence is the behaviour being
fixed. `ASSERT_DISCARD_STALE_ARTIFACTS=1` restores the previous delete.

Artifacts with no stamp predate this and are still treated as compatible, so existing
runs keep working.

`core.artifact_cache` had its own `file_sha256` that read whole files into memory; both
now share one chunked implementation so the cache digest and the provenance digest cannot
drift apart.

## Risk and rollback

### DIFF
--- a/assert_ai/core/artifact_cache.py
+++ b/assert_ai/core/artifact_cache.py
@@ -44,6 +44,7 @@ from pathlib import Path
 from typing import Any
 
 from assert_ai.core.io import PROMPTS_DIR, write_json
+from assert_ai.core.io import file_sha256 as _file_sha256
 
 
 log = logging.getLogger(__name__)
@@ -652,8 +653,10 @@ def hash_payload(payload: Any) -> str:
     return hashlib.sha256(_canonical_json(payload).encode("utf-8")).hexdigest()
 
 
-def file_sha256(path: Path) -> str:
-    return hashlib.sha256(path.read_bytes()).hexdigest()
+# Re-exported so existing callers keep working. The implementation lives in
+# core.io, reads in chunks rather than loading the whole file, and is shared
+# with the artifact provenance sidecars so both compute the digest identically.
+file_sha256 = _file_sha256
 
 
 def _stage_descriptor(
--- a/assert_ai/core/artifact_cache.py
+++ b/assert_ai/core/artifact_cache.py
@@ -293,7 +293,7 @@ def activate_latest_artifacts(ctx: dict[str, Any]) -> None:
             else resolved_metadata_path
         )
         metadata = _load_json_object(metadata_path)
-        if metadata and _metadata_outputs_exist(stage_name, artifact_dir, metadata):
+        if metadata and _metadata_outputs_valid(stage_name, artifact_dir, metadata):
             output_paths = _metadata_output_paths(stage_name, artifact_dir, metadata)
             # If the original ref's path entries pointed at locations that no
             # longer exist, rebuild the ref with the resolved on-disk paths so
@@ -808,7 +808,7 @@ def _latest_matching_metadata(
             continue
         hashes = metadata.get("hashes")
         if isinstance(hashes, dict) and hashes.get("input_hash") == input_hash:
-            if _metadata_outputs_exist(stage_name, version_dir, metadata):
+            if _metadata_outputs_valid(stage_name, version_dir, metadata):
                 matches.append((version_dir.name, metadata))
     return matches[-1] if matches else None
 
@@ -821,7 +821,7 @@ def _recover_latest_valid_version(
 
     for version_dir in reversed(_iter_version_dirs(stage_root)):
         metadata = _load_json_object(version_dir / ARTIFACT_METADATA_FILE)
-        if metadata and _metadata_outputs_exist(stage_name, version_dir, metadata):
+        if metadata and _metadata_outputs_valid(stage_name, version_dir, metadata):
             return version_dir.name, version_dir, metadata
     return None
 
@@ -839,14 +839,26 @@ def _is_safe_artifact_basename(filename: Any) -> bool:
     return True
 
 
-def _metadata_outputs_exist(
+def _metadata_outputs_valid(
     stage_name: str, version_dir: Path, metadata: dict[str, Any]
 ) -> bool:
-    """Return True iff every expected output file for ``stage_name`` exists.
+    """Return True iff every expected output for ``stage_name`` exists and is intact.
 
     Uses the merged path map from :func:`_metadata_output_paths` so a partial
     or missing ``metadata['files']`` cannot trick the cache into activating a
     half-written artifact (or one that is missing the primary output file).
+
+    Existence alone is not enough. ``finalize_artifact_plan`` records a sha256
+    for every output in ``file_hashes``, but nothing compared them on the way
+    back in, so a cached artifact altered after it was written — by an
+    interrupted run, a stray editor save, or anyone with write access to the
+    suite directory — was reused as though it were the computed result, and the
+    scores derived from it looked legitimate.
+
+    A mismatch is treated as a cache miss rather than an error: the stage
+    recomputes, which is the outcome the operator wanted anyway. Set
+    ``ASSERT_SKIP_CACHE_VERIFY=1`` to skip hashing if it is too slow on very
+    large artifacts, accepting that the cache is then unverified.
     """
 
     output_paths = _metadata_output_paths(stage_name, version_dir, metadata)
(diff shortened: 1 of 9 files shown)


changliu2 (changes requested):
The cache-hash check (commit 1) is a genuine improvement — it closes a case where a modified cached artifact was reused as if it were computed output. The preservation change is also straightforwardly better than `unlink()`.

Two things need to change before this merges, and one claim in the description needs to be walked back.

1. `verify_artifact_provenance()` and `check_artifact_schema()` have no callers in `assert_ai/` — only in `tests/`. The provenance sidecars are written by three stages and read by nothing. As it stands this commit records provenance; it does not verify it, and no stage can yet "tell whether its input came from the previous stage, a cache hit, or someone editing the file" as the description says. Either wire the check into the consumers (judge reading `inference_set.jsonl`, inference reading `test_set.jsonl`) behind a warn-only path, or retitle the commit and drop the SEC-021 closure until a follow-up does the wiring.

2. Sidecars go stale whenever the artifact they describe is mutated in place, and both mutation paths in this PR do exactly that:
   - `archive_artifact()` renames `scores.jsonl` aside but leaves `scores.jsonl.schema.json` behind, now describing bytes that are no longer there — and the preserved `.bak` carries no provenance at all, which defeats the point of preserving evidence.
   - The judge and inference resume paths `append_jsonl_row()` into an existing artifact and only rewrite the sidecar at the end of `run()`. A run that dies mid-append leaves a sidecar whose digest disagrees with the file. Once (1) is wired up, that surfaces as a false tamper report on a perfectly ordinary crash-and-resume.
   Sidecars should move with the file in `archive_artifact()`, and be deleted (or rewritten) before an artifact is appended to, not only after.

3. On the security framing: the digest lives next to the data, unsigned, and every fallback is permissive — missing `file_hashes` passes, a non-str recorded hash is skipped, `ASSERT_SKIP_CACHE_VERIFY=1` skips everything, and `artifact.json` itself is not protected. Anyone with write access to the suite directory (the threat named in the description and in SEC-005) can edit the artifact and then edit or delete the recorded hash beside it. This is solid corruption/accident detection and I'd merge it as that. It is not tamper resistance, and I don't think SEC-005/SEC-021 should be marked closed on it.

The tests are good where they exist — the tamper, skip-flag, and legacy-metadata cases are the right ones. What's missing is any test that exercises these paths through an actual stage: every new test calls the helpers directly, so the fact that nothing calls the verifiers is invisible to the suite.

**Inline notes**

`assert_ai/core/io.py:119` — `verify_artifact_provenance` (also `check_artifact_schema`, ~line 148). Dead code — callers only in `tests/test_artifact_preservation.py`. Fix: call `verify_artifact_provenance(inference_set_path)` in `judge.run_judge` after resolving its input, and on `test_set.jsonl` in `inference`, logging a warning on `False`. Or split into a follow-up and adjust the PR description.

`assert_ai/core/io.py:173` — `archive_artifact`. The `.schema.json` sidecar is neither moved with the file nor removed. Fix: if `path.with_name(path.name + ".schema.json")` exists, rename it to `backup.name + ".schema.json"` alongside the artifact.

`assert_ai/core/io.py:~200` — collision loop + `path.rename(backup)`. The `while backup.exists()` loop is not atomic with the rename, and on POSIX `rename` silently overwrites the target. Two runs archiving the same suite artifact within the same second can destroy one backup. Fix: use `os.link` + `unlink`, or `os.rename` guarded by `O_EXCL` creation of the target, and retry on collision rather than pre-checking.

`assert_ai/stages/judge.py:523` and `assert_ai/stages/inference.py:1383`. The sidecar is written only after `run()` completes, while the resume paths append to the existing artifact in place. A crash mid-append leaves a sidecar digest that no longer matches the file. Fix: delete the sidecar before entering the append/resume path.

`assert_ai/core/artifact_cache.py:~875` — `_metadata_outputs_valid`. Verification covers the version-dir copy, but `refresh_compatibility_files` copies outputs to suite-root working copies, and `_is_local_edit` deliberately preserves a suite-root file whose contents differ. Consumers that read the suite root directly still read an unverified — specifically known-divergent — file. Fix: at minimum log the divergence; ideally verify the suite-root copy at read time.

`assert_ai/core/artifact_cache.py:~880` — `ASSERT_SKIP_CACHE_VERIFY` and the `file_hashes`-missing fallback. Both are unauthenticated downgrades; `artifact.json` is not integrity-protected. Fix: log at WARNING (not silently return `True`) when verification is skipped for either reason.

`assert_ai/core/io.py:~155` — `check_artifact_schema`. Returns `False` for an artifact stamped with an older version, but `True` for an unstamped (genuinely older) artifact — same situation, opposite answer, and the docstring says the return value means "safe to read." Fix: return `True` for older-but-known versions and warn, or document the return as "matches this build's version."

`assert_ai/stages/systematize.py:131` — `taxonomy_json.setdefault("schema_version", ...)`. Injects a key into the model's structured output before writing. `TAXONOMY_SCHEMA` declares `additionalProperties: False`, so the written file no longer validates against the schema it was generated under; it also invalidates every downstream cache entry once, worth calling out under "Risk and rollback." Fix: write the version into the sidecar as with the JSONL artifacts, keeping the artifact byte-identical to the schema it was generated against.

`assert_ai/stages/test_set.py:1180` — `write_artifact_schema(out_path, artifact="test_set")`. No `produced_by`, so `test_set.jsonl` gets a schema stamp but no stage/version/run_id — inconsistent with inference and judge. It's also the one provenance-stamped artifact the cache copies between directories, and the sidecar is not in `_OUTPUT_FILES`, so it is not copied into the version dir and is not restored by `refresh_compatibility_files`. Fix: pass `produced_by`, and either include the sidecar in `_OUTPUT_FILES` for `test_set` or drop the sidecar during cache activation rather than leaving a stale one.

**Verdict: Request Changes** — the cache-integrity commit is worth merging, but the provenance commit ships verifiers that nothing calls and sidecars that go stale on the two paths this PR itself introduces, so the headline claim ("attribute") isn't yet true in running code.

**Must fix before merge**
1. Wire up `verify_artifact_provenance` / `check_artifact_schema`, or descope the commit and the SEC-021 closure claim.
2. Move/remove the sidecar in `archive_artifact`.
3. Invalidate the sidecar before append/resume mutates an artifact.
4. Correct the SEC-005/SEC-021 framing in the description — this is corruption detection, not tamper resistance.

**Nice to have**
- Atomic backup naming.
- Warn on skipped/unverifiable cache verification.
- `check_artifact_schema` older-version return-value inconsistency.
- Taxonomy `schema_version` injection vs. `additionalProperties: False`, plus a note about the one-time cache invalidation.
- `produced_by` and cache handling for the `test_set` sidecar.
- One end-to-end test through a stage rather than only direct helper calls.

Issue #291, opened 2026-07-30
[Bug]: Viewer defaults its headline and A/B compare axis to the coupled policy_violation metric


ango10:
Summary

- The results viewer uses the built-in policy_violation dimension as its default headline metric and default A/B comparison axis. policy_violation ORs over all relevant behavior nodes (permissible + non-permissible), so it couples the two failure modes into one number and can't be used to attribute A/B deltas. PR #276 already computes the decoupled split. We should repoint the headline, the compare default, and run ranking to the two halves of that split:
- policy_violation_not_permissible: harm only (non-permissible nodes)
- policy_violation_permissible: permissible behavior violated

Why policy_violation is the wrong default

- policy_violation  (rubric: "true = At least one relevant behavior is violated / false = No relevant behavior is violated", assert_ai/core/judge.py:83-88 ) is true when any relevant behavior is violated, permissible or not. So a governed run that reduces harm but increases permissible violations can leave policy_violation flat (or even rising) masking both effects. It mixes the two things you're trying to separate.

ango10:
Resolved with PR #295.

PR #295, opened 2026-08-01
feat(viewer): headline policy violations split by behavior

ango10:
## Summary

Make the permissibility-split policy violation metrics from PR #276 the headline pair throughout the viewer and CLI, and retire the metrics they supersede, so an ACS before/after separates harm reduction from over-restriction instead of reporting one blended rate. Issue: #291.

## Motivation / linked issue

Relates to PR #276, which computed the split into run artifacts but left the viewer headlining policy_violation, which is the single number blending two opposite failure modes: engaging with behavior the target must refuse (harm) vs. mishandling behavior it was allowed to handle. ACS moves those in different directions, so the blend hides the signal.

Once the split is the headline, keeping policy_violation and overrefusal on the same surfaces is redundant and misleading: policy_violation is the union of the two split buckets and overrefusal is a subset of the permissible bucket (refusal only). This PR also adopts the terminology the team aligned on: "Impermissible behavior violated" and "Permissible behavior violated", replacing "Harm (non-permissible)".

## Changes

Surfacing the split (#276 follow-through)
- Add viewer/src/lib/permissibility.ts: A single source of truth for the derivation, headline ordering, and visibility rules.
- Register policy_violation_not_permissible and policy_violation_permissible as built-in dimensions.
- Project the split onto verdict.dimensions during row normalization so grouping, filters, outcomes plots, csv exports, the drawer, and compare pick it up through the existing dimension pipeline instead of per-surface changes.

Retiring the superseded pair
- Add visibleMetricNames() and apply it to every display surface: run page, suite list, compare, export, and the outcome-plot dimension selector.
- Switch the assert-ai results list and results status tables to the split columns.
- Rename "Harm (non-permissible)" to "Impermissible behavior violated" and align the surrounding badges, subtitle, and footnote copy.
- Filter the drawer's node-judgement drill-down per permissibility bucket.
- Update both node-based test harnesses for the new module.

## Testing

- pytest tests/
- npm run check, npm run build
- Verified the derived dimension reproduces #276's artifact fields
- Manually exercised both paths in the viewer: split cards on runs with a taxonomy, and the fallback card on temporary artifacts copy with the taxonomy stripped.

## Checklist

image
image
image

### DIFF
--- a/assert_ai/cli.py
+++ b/assert_ai/cli.py
@@ -137,6 +137,12 @@ def _fmt_percent(value: Optional[float]) -> str:
     return f"{value * 100:.1f}%"
 
 
+_PERMISSIBILITY_SPLIT_RATE_KEYS = (
+    "not_permissible_policy_violation_rate",
+    "permissible_policy_violation_rate",
+)
+
+
 def _has_permissibility_split(*metric_sets: Any) -> bool:
     """True when any of ``metric_sets`` reports the permissibility split.
 
@@ -145,10 +151,16 @@ def _has_permissibility_split(*metric_sets: Any) -> bool:
     of permissible violations) on display surfaces. Runs without a behavior
     taxonomy -- including quality suites that repurpose ``policy_violation`` for
     non-safety failures -- have no split and keep reporting the original pair.
+
+    Detection keys off presence rather than a non-null rate: a bucket whose rate
+    is ``None`` was still computed, it just had no applicable rows. An
+    all-permissible taxonomy yields a real permissible rate alongside a ``None``
+    impermissible rate, and reading that ``None`` as "no split" would drop the
+    run back to the superseded pair while the viewer showed the split.
     """
     return any(
         isinstance(metrics, dict)
-        and metrics.get("not_permissible_policy_violation_rate") is not None
+        and any(key in metrics for key in _PERMISSIBILITY_SPLIT_RATE_KEYS)
         for metrics in metric_sets
     )
 
--- a/assert_ai/cli.py
+++ b/assert_ai/cli.py
@@ -137,6 +137,54 @@ def _fmt_percent(value: Optional[float]) -> str:
     return f"{value * 100:.1f}%"
 
 
+def _has_permissibility_split(*metric_sets: Any) -> bool:
+    """True when any of ``metric_sets`` reports the permissibility split.
+
+    The split supersedes ``policy_violation`` (which unions permissible and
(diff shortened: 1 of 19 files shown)


ango10:
Do not approve yet. Making changes per 08/03 standup.

- Align on naming convention: Impermissible behavior violated and Permissible behavior violated.
- Get rid of Policy Violation and Overrefusal throughout the viewer.

Update 08/03/2026 2:32PM PST: Changes complete. Waiting for approval.

AaronAspinwall123:
@ango10 please add some screenshots to the description if you update the UI

ango10:
Screenshots added.

changliu2:
Reviewed the diff end to end (permissibility.ts, labels.ts, cli.py, the run/compare/export surfaces). The engineering is solid — deriving the split from `node_judgments` + taxonomy rather than re-judging is the right call, and `visibleMetricNames` as a single choke point for the retire-superseded-metrics logic is clean.

One concern before approving, on the terminology itself (not the code): **""Permissible behavior violated"" and ""Impermissible behavior violated"" read as confusing to a new user**, and this PR makes them the *headline*, first-impression metric pair. The failure mode I'd worry about: `permissible` and `impermissible` differ by a 2-letter prefix, they're visually and phonetically close, and the phrase requires the reader to first parse ""permissible"" as a property of the *behavior category*, then realize the metric is about a *violation within* that category — that's two levels of indirection stacked before the number means anything. In a fast status-check scenario (glancing at a dashboard, or a screen-shared standup), I'd bet on real misreads between the two cards.

Concretely, what's actually being measured:
- `policy_violation_not_permissible` → did the agent **do something it should have refused**? (harm / safety failure)
- `policy_violation_permissible` → did the agent **mishandle something it was allowed to do**? (quality / helpfulness failure, the over-refusal-shaped bucket)

Those are two different *axes* (safety vs. quality), not two flavors of the same thing — the current naming doesn't signal that they're orthogonal, it just makes them look like a minimal-pair contrast a reader has to parse carefully to distinguish. I've dropped a few alternative label sets below for the team to consider — happy to pair on this before merge if useful. Nothing here blocks the underlying derivation/plumbing, which I think is right; this is purely about what we put in the headline UI copy.

AaronAspinwall123:
Two things I might look at is change  _has_permissibility_split()  to detect key presence, not a non-null impermissible rate, and add a regression test for an all-permissible run

AaronAspinwall123 (approved):

PR #296, opened 2026-08-03
chore(library): unbundle the three scenario specs into atomic behaviors

changliu2:
## Summary

Finishes the remaining branch-specific cleanup after #293 and syncs the branch with current `main`.

Against current `main`, this PR:

- keeps the travel-planner benchmark config scoped to the atomic `explicit_constraint_violation_failures` preset;
- keeps the canonical `travel_planner_benchmark` scenario context behavior-neutral and reusable across its listed atomic behaviors;
- updates the benchmark README to point to that canonical scenario instead of a bundled behavior document;
- deletes the orphaned `examples/benchmark/travel_planner_benchmark.md` bundle;
- preserves exact normalized parity between the identity-verification library preset and its Markdown reference; and
- adds a focused regression test proving the benchmark config reuses the canonical scenario context while selecting one atomic behavior.

## Review boundary

Current diff against `origin/main` (`da3a0bc`): **7 files changed, 29 additions, 62 deletions**.

The substantive benchmark change is the removal of the obsolete bundled spec and the explicit separation of reusable scenario context from the selected atomic behavior. Work already merged through #293 is not part of this remaining diff.

## Validation

- `python scripts/check_behavior_library.py` — 51/51 presets pass atomicity and parity checks.
- `python -m pytest tests/test_library_loader.py tests/test_library_e2e.py -q` — 90 passed, 803 subtests passed.
- Benchmark config loads successfully and has exact context parity with the canonical `travel_planner_benchmark` scenario.
- No repository references to `examples/benchmark/travel_planner_benchmark.md` remain.


examples/benchmark/eval_config.yaml
```diff
@@ -1,67 +1,18 @@
...
-      itinerary loses the "non-stop" qualifier).
-    - Returning vague, non-actionable recommendations ("there are several
-      good hotels in the area") when the user asked for concrete options
-      with names, prices, and links.
-    - Failing to surface a constraint conflict the tools clearly returned
-      (e.g., `validate_budget` flagged the plan as over-budget but the
-      itinerary doesn't mention it).
-context: 'The target is a multi-agent LangGraph travel planner with tool servers:
-
+  preset: explicit_constraint_violation_failures
```
ahmedmagooda, line 4: @changliu2 
So, this is essentially an example since we are moving to running multiple atomic evaluations for each scenario, correct ?

changliu2:
Correct. This config is now a single-atomic-behavior example, not the full benchmark. The old bundle covered 6 mechanisms in one config; those are now 6 separate atomic presets in `assert_ai/library/behaviors/` (`explicit_constraint_violation_failures`, `output_internal_consistency_failures`, `unit_conversion_failures`, `tool_call_turn_protocol_failures`, `actionability_failures`, `procedure_adherence_failures`). To reproduce full benchmark coverage for the `travel_planner_benchmark` scenario, you'd run one `eval_config.yaml` per atomic preset (each pointing `context:` at the same scenario) and aggregate results -- same pattern as `examples/travel_planner_langgraph/behaviors/*.yaml`, which already does this (one file per behavior: `constraints.yaml`, `grounding.yaml`, `tool-selection.yaml`, etc.). Happy to add a `benchmark/README.md` note showing the full run-all-presets sequence if that'd help clarify.

ahmedmagooda:
@changliu2 i think yes, adding the readme would be more clear. Other than that i think everything looks good.

changliu2:
@ahmedmagooda Pushed a follow-up commit (`83913da`) addressing your comment above: made the behavior/scenario library obviously discoverable at every entry point a user hits before writing a behavior spec by hand:

- Top-level `README.md` -- new nav-bar link (📋 Behavior Library) alongside Get Started/Examples/CLI Reference, plus a "What you get with ASSERT" bullet
- `docs/README.md` -- new Behavior Library entry under Configuration, described as "start here before writing a behavior spec by hand"
- `docs/config/best-practices.md` -- callout added directly inside §8.D ("Use atomic behaviors"), pointing to the library right where atomicity is explained
- `examples/README.md` -- new "Reuse a behavior from the library -- check here first" section, placed before "Which example to start with," referencing the real `assert-ai library list --kind behavior` / `assert-ai library show <preset>` commands

Can you review and approve when you get a chance?

examples/travel_planner_langgraph/README.md
tangym, line 77: This scenario table still appears to describe the previous bundled configuration. Could this table be updated so it matches the current `eval_config.yaml`?

scripts/check_behavior_library.py
```diff
@@ -92,9 +94,44 @@ def main() -> int:
...
+        for name, doc in sorted(behaviors.items()):
+            path = md.get(name)
+            if path is None:
+                fail(name, f"library preset has no {SPECS.relative_to(ROOT).as_posix()} reference")
+                continue
+            a, b = words(path.read_text(encoding="utf-8")), words(doc.get("description") or "")
+            if a != b:
+                import difflib
+                r = difflib.SequenceMatcher(None, a, b).ratio()
+                if r < 0.98:
```
tangym: For longer specs, a changed sentence could still pass while pip users receive different content from the markdown reference. Also, because the parity validation is guarded by `if SPECS.is_dir()`, removing the reference directory would skip the check entirely. 

The practical risk is low because yaml is the source of truth and md is not used at runtime; deleting the entire directory would likely be noticed during review.

examples/travel_planner_langgraph/behaviors/constraints.yaml
```diff
@@ -0,0 +1,105 @@
...
+# Every file here shares the same `context:` (the same application) and
+# measures exactly one mechanism, so a verdict is attributable to a single
+# behavioral claim. See docs/config/best-practices.md section 8.D.
+#
+# Run one:  assert-ai run --config examples/travel_planner_langgraph/behaviors/<file>
+# Gate all: configs: examples/travel_planner_langgraph/behaviors/*.yaml
+
+behavior:
+  preset: explicit_constraint_violation_failures
+context: 'The target is a multi-agent LangGraph travel planner with tool servers:
```
tangym: Style/readability: all six sibling configs serialize `context` as a hard-wrapped single-quoted scalar with blank lines between the wrapped source lines. The rubric strings use the same pattern. The runtime risk is low. Can we use the same literal-block style as the eval_config.yaml for these multiline fields?

```yaml
context: |
  ...
rubric: |-
  true = ...
  false = ...
```

changliu2:
@tangym Addressed both comments:

1. **README scenario table** — was still describing the pre-unbundling behavior (blended quality/safety failures, 6 categories). Updated to reflect the atomic prompt_injection quickstart config, and clarified the sibling ehaviors/*.yaml cover the other six mechanisms.
2. **YAML style** — reformatted context:/
ubric: in all six sibling configs (plus val_config.yaml's rubric for consistency) to literal-block style (|/|-), matching the existing convention. Verified byte-for-byte semantic equivalence via a yaml.safe_load diff against the prior committed content — pure style change, no content drift.

Also tightened check_behavior_library.py's parity check per your note: the 98%-similarity tolerance could let real content drift through silently in a long spec, since words() already normalizes every expected source of formatting difference. Now requires an exact match, and hard-fails (rather than silently skipping) if the xamples/behavior_specs reference directory goes missing.

Re-verified: 51/51 presets pass (atomic + parity) under the tightened rule, 89/89 targeted tests pass, and all 7 edited configs still load/resolve via ssert_ai.config.load_config.

Ready for another look whenever you have time.

changliu2:
Closed out the remaining review gaps from an independent audit pass:

1. **Ahmed's requested benchmark README** — added `examples/benchmark/README.md`, explaining this is a throughput-scale variant of the flagship `travel_planner_langgraph` example (same target, same `explicit_constraint_violation_failures` preset already used by `behaviors/constraints.yaml`), not a new agent or behavior. Registered it in `examples/README.md`'s selection table and layout tree so it's actually discoverable.
2. **Yeming's YAML-style comment** — one sibling config's rubric (`eval_config.yaml`'s `overrefusal`) was missed in the prior pass and still had the hard-wrapped single-quoted-scalar style; now literal-block like every other rubric/context field in the example. Verified byte-for-byte semantic equivalence via a `yaml.safe_load` diff.
3. **Yeming's `DeprecationWarning` comment on #293** — fixed on #293's branch directly (this branch is stacked on it, so it's included here too via merge).

51/51 presets clean, 89/89 targeted tests pass.

@tangym @ahmedmagooda — ready for another look whenever you have time.

jakepresent (changes requested):
Two things need tightening before this merges:

1. The post-#293 merge result leaves `examples/benchmark/travel_planner_benchmark.md` untouched. It still describes the old bundled behavior under `## Quality failures` (tool selection, verification, grounding, explicit constraints, unit conversion, internal consistency, completeness, actionability, and observation handling), while this PR only removes the README link to it. Nothing in the repository references the file afterward, so the benchmark directory still contains an orphaned bundled spec that contradicts the atomic pattern this PR is meant to establish. Please delete it or rewrite/reclassify it so the example directory no longer carries the obsolete bundle.

2. Please sync the branch with current `main` and update the PR summary to describe the remaining change. I simulated merging current `main` (`89c217a`) with this head (`07eeb0e`): the actual result is six changed files, 29 additions, and 18 deletions because most of the advertised 15-file/10-behavior work already landed through #293. The current Files changed view and body still present already-merged work as #296's diff, which makes the remaining review boundary unclear.

Validation on the exact synthetic merge: behavior-library checker passed (51 presets), focused library tests passed (90 tests and 803 subtests), and the benchmark config loaded with exact scenario-context parity.

changliu2:
@jakepresent Addressed the remaining exact-head blocker and synced the branch with current `main`.

- Deleted the orphaned `examples/benchmark/travel_planner_benchmark.md` behavior bundle.
- Kept `examples/benchmark/eval_config.yaml` scoped to one atomic behavior.
- Kept the canonical `travel_planner_benchmark` scenario context behavior-neutral and reusable.
- Updated the benchmark README and added a regression proving exact scenario-context reuse.

Validation: 51/51 library presets pass; 90 focused tests and 803 subtests pass.

Ready for re-review.

changliu2:
@jakepresent Synced the reviewed atomic scenario-unbundling fix onto current `main` after #336 (`81271d3`).

The merge was conflict-free. The obsolete bundled benchmark spec remains deleted, the benchmark config remains scoped to one atomic behavior, and the canonical scenario context remains behavior-neutral and reusable.

Validation still passes: 51/51 library presets; 90 focused tests and 803 subtests.

Ready for exact-head re-review.

tangym (approved):
My earlier feedback is addressed. The remaining changes are focused and all checks pass.

PR #305, opened 2026-08-06
fix(metrics): finish permissibility metric transition

jakepresent:
## Summary

Finish the user-facing permissibility-metric transition while preserving the legacy backend contract:

- show both permissible and impermissible violation rates for prompt and scenario runs
- make run detail, within-suite compare, cross-suite compare, and the post-run terminal headline use the split when the current taxonomy produces usable split data
- fall back to `policy_violation` / `overrefusal` when an older run's node judgments no longer match a regenerated suite taxonomy
- default comparisons to the impermissible split only when every compared run has split data, preserving the legacy delta table otherwise
- keep the legacy calculations in score rows, artifacts, JSON output, and explicit `--metric` comparisons for downstream compatibility
- carry the same temporary `arize-phoenix<19.18.0` cap as #279 because current 19.18–19.19 releases crash during pytest plugin import on supported Python 3.11
- remove redundant viewer projections, combined-policy counts, and special-case branches left dead after #295
- update the CLI reference for the data-aware compare default

This PR is based directly on `main`; it is no longer coupled to the skill/examples PR.

## Testing

- `pytest tests/ -x -q` → 1224 passed, 19 skipped, 474 subtests passed
- focused results/headline regressions → 17 passed
- `npm run check --prefix viewer` → 0 errors (6 pre-existing warnings)
- `npm run build --prefix viewer`
- `python -m compileall -q assert_ai tests/test_results.py tests/test_runner_headline.py`
- Ruff on changed CLI/results/tests plus fatal-error checks on `runner.py`
- `git diff --check`

## Compatibility

This does not remove the `policy_violation` or `overrefusal` judge dimensions or their artifact/JSON fields. Analysis, benchmark, and standalone export semantics remain unchanged pending a separate decision on whether they should report both split measures or retain the combined metric.

The prior CLI-only `permissible_overrefusal_rate` projection is intentionally removed from `results status --json`. It was not a persisted artifact field and had no in-repo consumer; the retained legacy compatibility fields are `policy_violation` and `overrefusal`.

### DIFF
--- a/assert_ai/cli.py
+++ b/assert_ai/cli.py
@@ -27,6 +27,7 @@ from assert_ai.results import (
     compute_dimension_summary,
     compute_policy_violation_by_permissibility,
     detect_dimensions,
+    has_permissibility_split_data,
 )
 from assert_ai.stages import STAGE_NAMES
 
@@ -254,7 +255,7 @@ def _visible_dimension_summaries(metrics: dict[str, Any]) -> list[tuple[str, dic
     dimensions = raw_dimensions if isinstance(raw_dimensions, dict) else {}
     visible: list[tuple[str, dict[str, Any]]] = []
 
-    if _has_permissibility_split(metrics):
+    if has_permissibility_split_data(metrics):
         for metric in (_POLICY_VIOLATION_NOT_PERMISSIBLE, _POLICY_VIOLATION_PERMISSIBLE):
             summary = metrics.get(_DERIVED_PERMISSIBILITY_SUMMARY_KEYS[metric])
             if isinstance(summary, dict):
@@ -484,12 +485,15 @@ def _dimension_rate(metrics: dict[str, Any], metric: str) -> float | None:
 def _resolve_compare_metric(metric: str | None, run_summaries: Iterable[dict[str, Any]]) -> str:
     if metric:
         return metric
-    for run_summary in run_summaries:
-        if _has_permissibility_split(
+    summaries = list(run_summaries)
+    if summaries and all(
+        has_permissibility_split_data(
             run_summary.get("prompt_metrics") or {},
             run_summary.get("scenario_metrics") or {},
-        ):
-            return _POLICY_VIOLATION_NOT_PERMISSIBLE
+        )
+        for run_summary in summaries
+    ):
+        return _POLICY_VIOLATION_NOT_PERMISSIBLE
     return DEFAULT_COMPARE_METRIC
 
 
@@ -497,7 +501,7 @@ def _available_compare_metrics(run_summaries: Iterable[dict[str, Any]]) -> set[s
     available: set[str] = set()
     for run_summary in run_summaries:
         available.update(_detect_dimensions(run_summary.get("prompt_rows") or []))
-        if _has_permissibility_split(
+        if has_permissibility_split_data(
             run_summary.get("prompt_metrics") or {},
             run_summary.get("scenario_metrics") or {},
         ):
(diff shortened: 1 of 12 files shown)


changliu2 (changes requested):
The five gaps this fixes are real — I diffed against `main` and confirmed each one (the missing "Scenario permissible violations" column, `results status` still listing the superseded pair while the viewer had already retired it, `--metric` unable to even name the derived split metrics, and `_log_run_headline` never loading the taxonomy so it couldn't compute the split). The viewer dead-code removals are clean: no remaining reads of `RunMetrics.counts` / `.policy_violation_rate` / `.overrefusal_rate`, and `aggregateRunViolationRate` had no callers. I reproduced your test results locally — 1216 passed / 22 skipped / 474 subtests, and `npm run check` is 0 errors / 6 pre-existing warnings.

Two blockers, both from the same root cause: the CLI detects "split is available" by key presence rather than data presence, and `_compute_prompt_metrics` emits those keys whenever the suite has any `behavior_categories` — regardless of whether the run's node judgments resolve against them.

**1. `results status` can render an empty dimensions table and hide real results.** With a fully-judged run whose `node_judgments` don't match the current taxonomy (routine whenever the taxonomy is regenerated between runs in a suite), both split buckets have `count == 0` but both keys are present, so `_visible_dimension_summaries` hides `policy_violation`/`overrefusal` and prints nothing else:

```
Total 1 - Scored 1 - Judge failure rate 0.0%
Dimension                        Summary  Scored  Distribution
Impermissible behavior violated  -        0       0 flagged / 0 pass
Permissible behavior violated    -        0       0 flagged / 0 pass
```

The underlying run has `policy_violation_rate: 1.0` and `overrefusal_rate: 1.0` — on `main` the user sees both. Worth noting the viewer doesn't have this bug, and `permissibility.ts` says why in its own docstring: `dropSupersededMetrics` requires "a list already narrowed to metrics carrying data." The CLI port dropped that precondition, so this adds a viewer/CLI asymmetry in a PR whose goal is to remove them. Could `_visible_dimension_summaries` gate on `policy_violation_on_{not_,}permissible` having `count > 0`?

**2. `results compare` silently switches metric and drops the delta table.** `_resolve_compare_metric` returns the split metric if any run reports the keys, so one split-capable run flips the default for the whole comparison. On the same fixture, `compare run-1 run-2` renders `-` for `run-1` and omits "Top behavior category deltas" entirely (every row returns `None` from `_row_metric_value`, so `first_map` is empty), where `--metric policy_violation` shows 100% for both runs. No warning is printed. Suggest requiring the split to carry data for every compared run before switching the default, or falling back with a one-line notice.

**Two non-blocking notes:**

- **Base branch.** This targets `ango10/assert-acs-skill-main` (#303), not `main`, so `gh pr checks` reports no checks at all — `build.yml` and `build-viewer.yml` are both `pull_request: branches: [main]`. A metrics-correctness fix is currently unvalidated by CI and coupled to an ~80-commit examples PR. The diff is self-contained; could it be rebased onto `main` and retargeted?
- **`permissible_overrefusal_rate`** is dropped from `results status --json`, which reads against "keep the legacy calculations in ... JSON output" in the description. I found no in-repo consumer (the export script computes its own), so this is probably fine — but since `--json` consumers are the ones we can't grep for, please either restore it or call the removal out explicitly.

Separately, five agent-instruction files still tell agents the viewer renders "Harm (non-permissible)" when the real label is "Impermissible behavior violated" (`.github/prompts/run-assert-eval.prompt.md:112,149`, `.claude/skills/run-assert-eval/SKILL.md:272,340`, `workflows/govern-and-remeasure.md:108`, `workflows/measure-clarity-failures.md:177`, `.cursor/rules/assert.mdc:148,186`). Those come from the base branch, so #303 may be the better place — but this is the PR claiming to finish the transition.

jakepresent:
Addressed both blockers and moved the PR directly onto `main`.

- Split metrics now replace the legacy pair only when the current taxonomy produces usable rows. Older runs whose node judgments no longer match a regenerated taxonomy keep their real `policy_violation` / `overrefusal` output instead of showing empty split tables.
- The default comparison switches to the impermissible split only when every compared run has split data. Mixed old/new runs fall back to `policy_violation`, including the behavior-category delta table.
- Added command-level regressions for run detail, suite listing, within-suite compare, one-sided taxonomies, and the post-run headline.
- Clarified in the PR body that the removed `permissible_overrefusal_rate` value was a CLI-only projection rather than a persisted compatibility field.

Final local gates: 1,224 tests passed with 19 skipped and 474 subtests; viewer check/build passed; focused regressions are 17/17.

jakepresent:
The first retargeted Tier 1 run hit the same upstream Phoenix 19.18–19.19 Python 3.11 import crash already fixed on #279. I carried that exact temporary `<19.18.0` dependency cap onto this branch as well; all other checks in the failed run were green, and fresh CI is running.

changliu2 (approved):
Thanks — I rechecked `0c4a401`. Both blockers from my prior review are closed: split metrics replace the legacy pair only when they carry usable data, and the default comparison switches only when every run has usable split data; mixed or stale-taxonomy runs fall back to `policy_violation`. The targeted regressions cover those paths, and the current check rollup is green. No remaining issues from me.

PR #314, opened 2026-08-12
fix(init): emit YAML block scalars for multi-line strings and non-ASCII

tangym:
## Summary

`assert-ai init` currently calls `yaml.dump(...)` with `default_flow_style=False` but without `allow_unicode=True` and without a block-scalar representer, so long strings emit as **double-quoted flow scalars with `\n` and `\uXXXX` escapes**. Behavior YAMLs land in customer repos and get read/edited by humans, so the current output (see [example on wire-assert-ci demo](https://github.com/responsibleai/assert-ai-action/tree/main/eval-demo-repo) — files with `description: "line one\nline two"` and `\u2014`) is nearly unreadable and diffs poorly.

## What changed

Consolidates all YAML emission through a new `assert_ai.core.yaml_io.dump_yaml()`:

- **Block-scalar for multi-line strings** — any string containing `\n` renders as `|` literal block, so `behavior.description`, `context`, judge dimension descriptions, and rubrics are readable.
- **`allow_unicode=True`** — em-dashes, curly quotes, CJK stay verbatim instead of becoming `\uXXXX`.
- Keeps `default_flow_style=False` + `sort_keys=False` so author key order is preserved.

Wires the helper into every existing emitter:

| File | Change |
|---|---|
| `assert_ai/core/yaml_io.py` | New shared helper + `BlockStyleDumper` |
| `assert_ai/init/_emit.py` | Atomic writer → `dump_yaml(data)` |
| `assert_ai/init/_context.py` | Preset dumps in LLM-facing prompts → `dump_yaml(preset)` |
| `assert_ai/init/_design_agent.py` | `_normalize_yaml` used for interactive drafts + `eval.draft.yaml` |
| `assert_ai/integrations/acs/eval_config.py` | Delete duplicate `_EvalConfigDumper` / `_represent_str`; keep `_dump_eval_config_yaml` as thin wrapper |
| `assert_ai/cli.py` | `assert-ai preset show` uses the same helper |

## Tests

`tests/test_yaml_io.py` — 6 new focused tests:

- multi-line string uses `|`
- single-line stays plain
- non-ASCII emitted verbatim
- multi-line + non-ASCII combined
- key-order preservation
- nested-mapping block style

All 6 pass. Pre-existing `tests/test_acs_eval_config.py` (6 tests) still passes — the ACS eval-config generator is now backed by the same shared helper.

## Sanity check

Ran the fixed `emit_config` on a sample with multi-line rubrics and em-dashes:

```yaml
behavior:
  name: accurate_attribution
  description: |
    Multi-line description with an em—dash.
    Line two.
```

No `\n` escapes, no `\u2014`, em-dashes verbatim, `|` block style throughout.

## Notes

- The helper subclasses `yaml.SafeDumper`, so all safety guarantees of `safe_dump` are preserved.
- Existing files that were already flow-style don't need migration; PyYAML will read them fine — new files just emit better.

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -40,22 +40,14 @@ jobs:
           cache-dependency-path: viewer/package-lock.json
 
       - name: Install Python dependencies
-        # Deliberately does NOT install the `otel` extra. `otel` pulls in
-        # `arize-phoenix`, which registers a `pytest11` entry point whose
-        # import chain runs `phoenix.trace.dsl.filter` at pytest collection
-        # time. Recent phoenix releases (19.18+) declare a frozen dataclass
-        # with a `mappingproxy` default there, which Python 3.11's dataclass
-        # machinery rejects, taking down every test job that has phoenix
-        # installed — including our own unit tests, which don't use phoenix.
-        #
-        # The base `dependencies` already ship `opentelemetry-api` / `-sdk`
-        # (all the `otel` demos need at import time), and
-        # `assert_ai.auto_trace.enable(...)` guards its `phoenix.otel`
-        # import behind a collector-availability probe, so smoke tests that
-        # instantiate the incident-triage agent still work without phoenix
-        # in the install set.
+        # The `otel` extra is required because tests/test_incident_triage_smoke.py
+        # imports examples/incident_triage_agent/agent.py, which imports
+        # opentelemetry at module load time (target.trace.backend: otel). Demo
+        # examples like that one ship as part of the unit-test surface, so the
+        # CI install set has to include the demo's optional extras even though
+        # core ASSERT doesn't need them.
         run: |
-          python -m pip install -e ".[dev]"
+          python -m pip install -e ".[dev,otel]"
 
       - name: Install viewer npm dependencies
         # tests/test_viewer_*.py shell out to `node` against viewer TypeScript
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -40,14 +40,22 @@ jobs:
           cache-dependency-path: viewer/package-lock.json
 
       - name: Install Python dependencies
-        # The `otel` extra is required because tests/test_incident_triage_smoke.py
-        # imports examples/incident_triage_agent/agent.py, which imports
-        # opentelemetry at module load time (target.trace.backend: otel). Demo
-        # examples like that one ship as part of the unit-test surface, so the
-        # CI install set has to include the demo's optional extras even though
-        # core ASSERT doesn't need them.
+        # Deliberately does NOT install the `otel` extra. `otel` pulls in
+        # `arize-phoenix`, which registers a `pytest11` entry point whose
+        # import chain runs `phoenix.trace.dsl.filter` at pytest collection
+        # time. Recent phoenix releases (19.18+) declare a frozen dataclass
+        # with a `mappingproxy` default there, which Python 3.11's dataclass
+        # machinery rejects, taking down every test job that has phoenix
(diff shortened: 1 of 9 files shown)


jakepresent (dismissed):
Verified on the current main-based head. The shared SafeDumper preserves round-trip values, key order, Unicode, and literal block scalars across init, preset display, and ACS config emission. I removed the unrelated CI workaround from the PR so it keeps the release's Phoenix dependency gate and remains scoped to YAML output. Local verification: 1,230 passed, 19 skipped, 474 subtests; viewer check/build passes; the complete GitHub build, install, CodeQL, and Tier 1 matrix is green. Merge-clean.

changliu2 (changes requested):
Three serialization/output regressions block this as written:

1. `allow_unicode=True` makes U+0085 (NEXT LINE) non-round-trippable. `{"x": "a\x85b"}` reloads as `"a b"`; the prior escaped `\N` form preserved the value. Please force double-quoted style for strings containing U+0085 and add single-/multiline round-trip tests.
2. `click.echo(dump_yaml(data).rstrip())` can trim user data now that Unicode whitespace may be emitted literally. A trailing NBSP is removed, and an NBSP-only scalar can reload as `null`. Please preserve the serialized document exactly, e.g. `click.echo(dump_yaml(data), nl=False)`.
3. Raw Unicode can raise `UnicodeEncodeError` on redirected Windows `cp1252` stdout—for example a preset containing `日本語`. Please make the output path UTF-8-safe or fall back to escaped scalars for unencodable streams, with a subprocess regression.

The focused tests pass, but these cases can corrupt data or make `assert-ai init` fail on supported Windows output paths.

tangym (commented):
Thanks for the review — all three findings addressed in [`0d9b4255`](https://github.com/responsibleai/ASSERT/commit/0d9b4255) and [`c03144d2`](https://github.com/responsibleai/ASSERT/commit/c03144d2).

**1 — U+0085 / U+2028 / U+2029 round trip.** Added a guard in `_represent_str` that forces double-quoted style when a string contains any of those three code points, so PyYAML falls back to `\N` / `\L` / `\P` escapes and the value survives the round trip. All other strings keep the current behavior (block scalar for multi-line, plain for single-line, verbatim non-ASCII). New tests cover single-line and multi-line round trips for all three.

**2 — NBSP dropped by `.rstrip()`.** Root cause was `.rstrip()` mutating a serialized YAML doc before print; the NBSP loss is one symptom of the class. Removed entirely in favor of writing the YAML byte-exact through `sys.stdout.buffer` — no rstrip, no re-encode. Also collapses (3) since the same path handles both.

**3 — `UnicodeEncodeError` on cp1252 stdout.** New `_write_stdout_utf8` helper writes UTF-8 bytes through `sys.stdout.buffer`, with a fallback to `click.echo` for text streams without a binary layer (Click's `CliRunner`). Deterministic output across platforms — the bytes a customer gets from `library show > out.yaml` on Windows now match what `assert-ai init` writes on Linux. Regression test spawns `assert-ai library show travel_planner` (bundled preset containing em-dashes) under `PYTHONIOENCODING=ascii` and asserts the em-dash survives to UTF-8 stdout.

Full suite green locally, including the ACS eval-config and init emit paths that share `dump_yaml`.

changliu2 (approved):
Re-reviewed at `4dc2a8e`. All three blockers are fixed: U+0085/U+2028/U+2029 are emitted in escaped double-quoted form and round-trip; serialized YAML is no longer trimmed; and `library show` writes UTF-8 bytes on redirected stdout with a text-stream fallback. The shared emitter and regression coverage are coherent, and required checks are green. Approving.

PR #317, opened 2026-08-14
feat(skill): run-assert-eval smoke-run 3 real test cases before full inference

### LINKED ISSUE #316: [Feature]: run-assert-eval smoke-run 3 real test cases before full inference
### Problem you're trying to solve

The `run-assert-eval` Skill generates a fresh eval_config.yaml per risk, so almost every config is unvalidated on first execution. Config and plumbing errors (wrong `callable`, missing credentials, target raising on first tool call, tool-schema mismatch, undeployed judge model) only surface once inference starts (after systemize and test_set have already run. A 25 prompt + 25 scenario suite takes a long time, and the governance workflow is an A/B, so every mistake costs two runs.

### Proposed solution or behavior

Add a smoke gate to the Skill workflow, between test-set generation and full
inference:

    Test set ready: 25 prompt / 25 scenario.
    Smoke test 3 prompt cases before the full run? [Y/n]

On yes, the Skill runs the same config file four times over, no derived config needed:

    # 1. upstream only -> full taxonomy + full 50-row test set
    assert-ai run --config C \
      --override inference.enabled=false --override judge.enabled=false

    # 2. slice rows where type == "prompt", take 3  (~3 lines of Python)

    # 3. smoke -> different run label, different input file
    assert-ai run --config C --override run=baseline-smoke \
      --override inference.test_set_path=artifacts/smoke/test_set_3.jsonl

    # 4. full run -> systematize + test_set are cache hits
    assert-ai run --config C

On failure at step 3, stop and surface the error rather than proceeding.

Why this is safe:

  - `pipeline.test_set` is never modified, so the full run's cache key is
    untouched and step 4 costs nothing extra upstream.
  - The 3 cases are real rows from the real test set, not a regenerated
    smaller design.
  - Artifacts live at `artifacts/results/<suite>/artifacts/`, a sibling of the
    run dirs and keyed by suite, so a distinct `run:` label is free, the same
    mechanism that already lets baseline and acs-governed share cached
    systematization.
  - Step 3 leaves systematize and test_set enabled: both are cache hits with
    zero LLM calls, and keeping them on means the judge still receives its
    taxonomy from context, so no `taxonomy_path` wiring is needed.

All three overrides are already supported and coerce correctly (verified against
`_apply_config_overrides`); explicit `inference.test_set_path` wins over context
(`stages/inference.py:1348`).

### Alternatives considered

A first-party `assert-ai run --smoke N` flag. Would help CI and non-Skill users, but adds no capability the Skill lacks, the primitives already exist. The conversational gate is a Skill concern anyway; a CLI flag can't ask. Worth revisiting only if the pattern proves out and other callers want it.

### Use case context

Target type: `callable` (e.g.
`examples.billing_support_agent.agent:chat_baseline`), with the same suites
re-pointed at a governed callable for the A/B. Applies to model and model+tools
targets equally.

Framework: custom Python agents over litellm; also targets in `examples/`.

Stage: prototyping / config authoring, not production.

Environment: Azure OpenAI via litellm. `azure/gpt-5.4-mini` targets, `azure/gpt-5.4` for systematize and judge

### Acknowledgements

- [x] I searched existing issues and discussions and did not find a duplicate.
- [x] I'd be willing to contribute a PR for this (optional).

ango10:
## Summary

Adds a smoke-run capability to the `run-assert-eval` skill: validate a config end-to-end on 3 real test cases before committing to a full 25 prompt + 25 scenario run. Implemented entirely with existing ASSERT primitives. No changes to `assert_ai/`.

## Motivation / linked issue

Resolves #316.

Plumbing errors: a wrong `target.callable`, missing credentials, a callable that raises on its first tool call, a tool-schema mismatch, an undeployed judge model, etc. only surface once inference starts, after systematize and test_set have already run. Today the cheapest way to find out is to pay for the whole suite.

## Changes

- **`smoke_slice.py` (new)**: slices the first N rows of a given kind out of a suite's already-generated test set to `artifacts/smoke/<suite>-<kind>-<n>.jsonl`, and prints a JSON summary. Stdlib-only (PyYAML only for the optional `--config` form). Three deliberate safety properties:
  - resolves the test set through `latest.json`, the pointer ASSERT itself maintains, rather than assuming `v0001`. Version dirs are allocated fresh on every cache miss.
  - copies raw line text, so sliced rows are byte-identical to rows the full run will score.
  - refuses to write inside the suite root, which could flaw the published `test_set.jsonl` and invalidate the very cache the smoke run exists to protect.
- **`tests/test_smoke_slice.py` (new)**: 27 tests; all build isolated suite trees under `tmp_path` and never touch the real `artifacts/results/`.
- **`workflows/measure-clarity-failures.md`**: new **Step 5a**, offered by default: generate artifacts with `inference.enabled=false --override judge.enabled=false`, slice 3 rows, run them under `run=<run>-smoke`, then gate. Documents where to read the result (`results status`, viewer URL), that a smoke run is a **run inside the suite and not a suite card**, and a "never substitute these" block covering the `sample_size` trap, the separate `run:` label, and the suite-root write ban.
- **`workflows/govern-and-remeasure.md`**: smoke the governed callable before the governed run, since that's where a wrong `agent_guarded` path is most likely and a failure wastes half an A/B.
- **`SKILL.md`, `.github/prompts/run-assert-eval.prompt.md`, `.cursor/rules/assert.mdc`**: the run step in all three surfaces the README requires kept aligned.
- **`README.md`, `SETUP-CHECKLIST.md`**: slicer reference section and two verification bullets.

Documentation-only outside the two new files; no `assert_ai/` code changed.

## Testing

- `pytest .claude/skills/run-assert-eval/tests/` → **56 passed** (29 existing + 27 new). This is exactly how CI invokes it; `regression.yml` already globs `.claude/skills/**` and that test directory, so the new suite is picked up with no workflow change.
- `pytest .claude/skills/run-assert-eval/tests/ tests/test_init_command.py` → **68 passed** after merging `main`, since the incoming #314 touches config emission.
- **Real end-to-end run** on `travel-langgraph-budget-overrun`: 3 prompt cases, inference 38.6s + judge 14.9s = 53.6s, 3 judge calls, judge failure rate 0.0%. Artifacts, `scores.jsonl`, `metrics.json` and all five `.viewer/` read-model files were produced, and the run is discoverable via both `assert-ai results status <suite> baseline-smoke` and the viewer run page.
- **Both documented `--override` commands validated against the real config loader** (`_apply_config_overrides` + `load_runtime_context`, no LLM calls): step 1 disables inference/judge as genuine booleans; step 3 keeps `suite_root` unchanged so systematize/test_set stay cache hits, routes the slice to `inference.test_set_path`, and leaves `test_set.prompt.sample_size` at 25.

## Checklist


ango10:
image
image

jakepresent (changes requested):
Two path-boundary issues need fixing before this helper ships.

1. `--suite` is treated as a filesystem path rather than a suite identifier. `build_slice()` does `(results_dir / suite).resolve()` without requiring the result to remain under `results_dir`, and the default output filename is derived from that same unsanitized value. I passed an absolute temporary directory as `suite`; the helper successfully read its `test_set.jsonl`. With the default output, an absolute or traversal value can also escape `artifacts/smoke/`. Please apply the same suite-identifier validation as ASSERT and require `suite_root` to stay within `results_dir` before opening anything. Add absolute-path and `..` regressions.

2. `resolve_results_dir()` does not actually match `assert_ai.config`. For a supported config containing `artifacts_root: artifacts` and `results_dir: artifacts/custom`, ASSERT resolves `<root>/artifacts/custom`, while the slicer resolves `<root>/artifacts/artifacts/custom`. The documented `--config` path then looks in the wrong tree. Please use the canonical config resolver or mirror its artifact-prefix handling and add this case to the tests.

The focused suite otherwise passed (`56 passed`), Ruff and compile checks passed, and the full repository suite passed after bootstrapping viewer dependencies (`1,365 passed, 25 skipped, 834 subtests`).

Coordination note: this branch and the optional-Clarity branch overlap in seven skill/workflow files. Land the optional-Clarity behavior first, then sync this branch and recheck the resolved instructions so neither feature is dropped during conflict resolution.

jakepresent (approved):
Re-checked at `0222d7e`. Both original blockers are addressed:

- `suite` is now validated with ASSERT's identifier contract before any read or default-output construction, and the resolved suite root is required to remain under `results_dir`; the absolute-path and traversal probes are rejected.
- `results_dir` now matches ASSERT's artifact-prefix behavior for default, custom, prefixed, renamed-root, and absolute paths.

The main sync also preserved both features across the overlapping skill surfaces: explicit user-supplied risks still win, and the smoke-run instructions remain present in the Claude, Copilot, and Cursor versions.

Verification: 39 slicer tests and 68 total skill tests passed; Ruff, compile, behavior-library parity, and diff checks passed; full repository suite passed (`1,366 passed, 24 skipped, 834 subtests`). Approving.

PR #319, opened 2026-08-14
fix(skill): make Clarity risk discovery optional, not a required gate.

### LINKED ISSUE #318: [Bug]: Clarity integration should not be mandatory
### What happened?

`.clarity-protocol/` (or a live Clarity MCP session) is a hard precondition for running the `run-assert-eval` skill. There is no supported way to bring your own risks, so a user who already knows what they want to evaluate cannot proceed.

This is explicit and intentional in the skill text:

- `SKILL.md:32`: a dedicated section titled **"Clarity is required for Run mode — no non-Clarity fallback"**
- `SKILL.md:34-38`: "Risks that seed an eval **MUST** come from Clarity … Do **not** substitute a plain-language description"
- `SKILL.md:67-73`: the Clarity MCP server is listed as a required *precondition*; if the tools are missing the skill is told to divert the user into `SETUP-CHECKLIST.md`
- `SKILL.md:428` (Guardrails): "if the MCP tools can't be made available, **stop and help fix it**"

The same two blocks are mirrored verbatim in `.github/prompts/run-assert-eval.prompt.md:43,193` and `.cursor/rules/assert.mdc:65,240`, so the gate is identical in Claude Code, Copilot, and Cursor.

Neither documented entry mode is an escape hatch (`SKILL.md:20-30`): **Run mode** requires Clarity, and **Results Q&A mode** requires judged artifacts to already exist under `artifacts/results/<suite>/<run>/`. A new user with an agent and a clear risk in mind matches neither.

The practical effect is that a hard dependency on a **separate, non-shipped tool** blocks the entire onboarding path for ASSERT's own skill.

### What did you expect to happen?

Clarity should be the strongly recommended default, not a gate.

When no `.clarity-protocol/` exists, the skill should offer a choice rather than stopping:

1. **Run Clarity discovery** (recommended, current behavior): for users who don't yet know their risks.
2. **Bring your own risks**: accept a plain-language description, a PRD, a design doc, a threat model, or an existing test plan.

On path 2 the skill should still drive the best-practice experience it already encodes, rather than degrading into a freeform chat: elicit an **atomic** behavior (one risk, one config), derive stratification dimensions and elicitation variants, pin the systematize/judge models, ask for `sample_size`, set `max_turns`, and produce exactly the same config shape Clarity-sourced risks produce. The quality bar the skill enforces comes from Steps 3-6, which are risk-source agnostic.

Clarity's advantage is *discovering* risks the user hasn't thought of. That advantage is irrelevant when the user already has them written down, and it shouldn't be a precondition for measurement.

The loop-closing steps that write back to `.clarity-protocol/` (`record_failure` / `record_suggestion`) should degrade gracefully to no-ops with a note, instead of being treated as required.

### Reproduction steps and minimal eval config

1. Open a repo with an agent but **no** `.clarity-protocol/` directory, in an IDE session where the `clarity-agent` MCP tools are not configured.
2. Prompt: `Help me evaluate my <travel planner> agent.`
3. Observe the skill refuses to proceed to config generation and instead routes into Clarity setup:

> "No `.clarity-protocol` exists yet, so this is a fresh Run mode. Let me inspect the existing travel planner example and finish reading the workflow.
>
> I have enough context on the skill workflow. No `.clarity-protocol/` exists, so this is a fresh Run mode: Clarity is the required risk-discovery front door. Let me set up a task list and gather the remaining preconditions."

4. Try to bypass it: e.g. `I already know the risk: the agent books itineraries over the stated budget. Skip Clarity.` The guardrail at `SKILL.md:428` directs the assistant to stop and fix the MCP setup instead.

**Minimal eval config: not applicable.**

### Version info (assert_ai version, Python version, OS)

`assert-ai` **0.1.0**, Python **3.12.10**, Windows 11 Enterprise

### Relevant logs or traceback

```shell

```

### Acknowledgements

- [x] I searched existing issues and did not find a duplicate.
- [x] I redacted secrets, credentials, and customer data from my report.
- [x] I included a minimal reproduction.

ango10:
## Summary

Makes Clarity risk discovery **optional** in the `run-assert-eval` skill. Users can seed an eval from their own description, PRD, threat model, or incident report instead of being blocked at a mandatory `.clarity-protocol/` gate.

## Motivation / linked issue

Resolves #318. 

The skill treated `.clarity-protocol/` as a hard precondition: with no protocol present it would stop and walk the user through MCP setup rather than proceed. That blocks anyone who already knows the risk they want measured, and anyone who can't run Clarity at all (it needs Python 3.12+ and an MCP-capable IDE). The root README stated the constraint outright. *"Risks always come from Clarity the skill won't let you seed an eval from an off-the-cuff description."*

Clarity remains the **recommended** path. It is no longer the only one.

## Changes

- **Replaced the gate with an explicit risk-source choice** in Step 1: **1a** Clarity discovery (recommended) or **1b** user-supplied risks. Both emit the same candidate-behavior contract, so every downstream step (triage, config generation, running, judging, governance) is risk-source agnostic.
- **Added a structured intake for Path B**: context, prohibited behavior, the permissible boundary, elicitation variants, and severity: the same fields `clarity_intake.py` already produces, so no downstream code changes were required.
- **Made Clarity write-backs conditional** on a protocol actually existing (`measure-clarity-failures.md` Step 8, `govern-and-remeasure.md` Step 7), and marked the measurement workflow's Step 1 (Parse) skippable on Path B.
- **Mirrored across all four skill surfaces** so behavior is identical in every IDE: `SKILL.md` (canonical), `.github/prompts/run-assert-eval.prompt.md`, `.cursor/rules/assert.mdc`, and `.github/copilot-instructions.md`.
- **Corrected the docs that asserted the requirement**: root `README.md` (including splitting onboarding, so ASSERT needs 3.11+ and only Clarity needs 3.12+/MCP), `AGENTS.md`, the skill `README.md`, and `SETUP-CHECKLIST.md` (now marked Optional, with a Path B verification item).

Unchanged by design: the mandatory **human triage gate**, the `.clarity-protocol/` **preservation STOP**, and the ban on the agent imitating Clarity's interview.

## Testing

- Repo-wide wording sweep: enumerated all 29 files mentioning Clarity across all file types, plus multiline greps (a gate word on the line following a Clarity mention) and indirect phrasings ("front door", "only source", "risk-discovery engine"). 
- Manual: with `.clarity-protocol/` absent, ran the skill against a domain with no eval configs and no result suite; confirmed it offers the risk-source choice instead of stopping to demand MCP setup.

## Checklist


ango10:
image
image

jakepresent (changes requested):
The optional path is mostly aligned with yesterday's decision, but the central choice contract still contradicts itself. The skill says the user chooses and that we must never silently pick a path, then Step 1 says an existing `.clarity-protocol/` is selected automatically with “no need to ask.” The same automatic default appears in the Copilot and Cursor mirrors.

That makes Path B unavailable in a common case: a developer explicitly asks to measure a known risk while an older protocol is present. The skill will silently switch to the protocol's risks, and the preservation gate does not help because it only runs before a fresh Clarity discovery. Please honor an explicit user-supplied risk regardless of whether a protocol exists. When intent is ambiguous, offer the existing protocol as the default but ask before selecting it. Mirror that behavior across all three surfaces.

Verification on this head: the Clarity wording sweep found no remaining mandatory-gate phrases, skill tests passed, behavior-library parity passed, and the full repository suite passed after bootstrapping viewer dependencies (`1,365 passed, 25 skipped, 834 subtests`).

jakepresent (approved):
Re-checked at `0bdc77a`. The original concern is addressed across the Claude, Copilot, and Cursor surfaces: an explicit user-supplied risk now always wins, an existing protocol changes only the recommended default, and ambiguous intent still requires the user to choose. The stale automatic-selection wording is gone.

Verification: 29 skill tests passed, behavior-library parity passed, compile and diff checks passed. Approving.

PR #324, opened 2026-08-19
fix(viewer): stop clipping eval results table and metric dropdown.

### LINKED ISSUE #322: [Bug]: Viewer `permissible_behavior_violated` is cut off
### What happened?

Image

### What did you expect to happen?

see image

### Reproduction steps and minimal eval config

1.
2.
3.

```yaml
# minimal eval_config.yaml
```

### Version info (assert_ai version, Python version, OS)

see image

### Relevant logs or traceback

```shell

```

### Acknowledgements

- [x] I searched existing issues and did not find a duplicate.
- [x] I redacted secrets, credentials, and customer data from my report.
- [x] I included a minimal reproduction.

ango10:
## Summary

Fixes the evaluation-results table on the suite page, where the right-hand columns were clipped rather than scrollable, and fixes the metric dropdown menus that the scroll-container fix then exposed as clipped.

## Motivation / linked issue

Closes #322.

The Total column was unreachable on **every** suite. Not just long metric names, and not just narrow monitors. The page container is Primer's `container-xl` (max 1280px), so usable width is ~1198px regardless of screen size, while the table rendered at 1331px. The wrapper used `overflow-hidden`, so that 133px simply disappeared with no scrollbar to recover it.

## Changes

- **`overflow-hidden` → `overflow-x-auto`** on the results-table wrapper, matching the pattern already used on the compare page. This is the safety net: no column can become unreachable again, whatever metrics a user selects.
- **Metric header labels now wrap inside a fixed 10rem column.** These were the main width driver. The `<th>` was declared `w-32`, but `PrimerDropdown`'s trigger is a fixed-height `white-space: nowrap` button, so `w-32` could never bind — "Impermissible behavior violated" alone forced the column to 285px. Wrapping brings the table to 1198px, exactly inside the container, without shortening any metric name. `labels.ts` is shared with other surfaces and is deliberately untouched.
- **Capped the run-target cell so its existing `truncate` applies.** The `truncate` class there was dead code: neither it nor `min-w-0` can bind inside an auto-layout `<td>`, which grows to max-content, so the full target string rendered every time. A `max-w` on an inner block element makes the truncation and its `title` tooltip work as originally intended.
- **`PrimerDropdown` menus now render `position: fixed`, anchored to the trigger's viewport rect.** Making the wrapper a scroll container exposed a second bug: an absolutely-positioned menu cannot escape an ancestor scroll context, so an open metric dropdown was cut off at 101px of overhang and its options were unreadable. It also inflated the wrapper's `scrollWidth` (1198 → 1300), producing a stray horizontal scrollbar whenever a menu was open. A fixed menu is neither clipped nor counted in an ancestor's scroll extent, so both symptoms resolve from one change.
- **Added the collision handling `PrimerDropdown` previously lacked**: horizontal flip to right-aligned, vertical flip above the trigger, viewport clamping, and repositioning on scroll/resize while open.

## Testing

- `npm run check`: 0 errors (6 warnings, all pre-existing in `routes/new/+page.svelte`).
- `npm run build`: passes.
- **Table layout**, headless browser, 5 suites × 3 viewports (1280/1440/1680): 15/15 have no unreachable column. Negative control against the pre-fix code: 15/15 failed, with 23–227px of overflow.
- **Dropdowns**, headless browser, 3 pages × 3 viewports: 15/15 render `position: fixed`, fully in-viewport, with unclipped option text. Negative control: all 15 were `absolute` before, 3 of them escaping the viewport entirely.
- Confirmed the stray horizontal scrollbar no longer appears while a menu is open.

## Checklist

### DIFF
--- a/viewer/src/lib/PrimerDropdown.svelte
+++ b/viewer/src/lib/PrimerDropdown.svelte
@@ -36,6 +36,47 @@
   let highlightedIndex = $state(-1);
   let triggerButton: HTMLButtonElement;
   let menuList: HTMLUListElement;
+  // The menu is rendered with `position: fixed` and anchored to the trigger's
+  // viewport rect so it can escape any ancestor with a scroll/clip context
+  // (e.g. the `overflow-x-auto` wrapper around the evaluation-results table).
+  // An absolutely-positioned menu is clipped by such an ancestor and also
+  // inflates its scrollWidth, which spawns a stray horizontal scrollbar.
+  let menuStyle = $state('');
+
+  const VIEWPORT_MARGIN = 8;
+  const TRIGGER_GAP = 4;
+  const MIN_MENU_WIDTH = 192; // matches the 12rem min-width in CSS
+
+  function positionMenu() {
+    if (!triggerButton || !menuList) return;
+    const anchor = triggerButton.getBoundingClientRect();
+    const menu = menuList.getBoundingClientRect();
+
+    // Prefer left-aligned to the trigger; flip to right-aligned when that would
+    // overflow the viewport, then clamp so the menu is never off-screen.
+    let left = anchor.left;
+    if (left + menu.width > window.innerWidth - VIEWPORT_MARGIN) {
+      left = anchor.right - menu.width;
+    }
+    left = Math.max(
+      VIEWPORT_MARGIN,
+      Math.min(left, window.innerWidth - menu.width - VIEWPORT_MARGIN)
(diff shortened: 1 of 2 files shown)


ango10:
before: Image

after:
image

ango10:
Changes have already been implemented in existing viewer dashboard found here. Metrics are no longer cutoff on dropdown: 
image

PR #326, opened 2026-08-20
fix(sandbox): keep egress evidence outside target mount

jakepresent:
## Problem

The stock sandbox wrote host-proxy network evidence under the same directory mounted read/write into the evaluated target. A target could overwrite or delete `egress.jsonl` before ASSERT consumed it.

## Fix

- keep the proxy-generated egress ledger in a sibling host-only `audit/` directory
- continue mounting only the target's declared `output/` directory
- document the host-owned network-evidence boundary
- add unit and real-Docker regressions proving a forged `/sandbox/output/egress.jsonl` cannot alter the host ledger

This PR fixes network-ledger mutability only. Moving tool-action capture out of the evaluated process is a separate architectural change.

## Verification

- Regression test was added first and failed against the previous path (`out/egress.jsonl` instead of host-only `audit/egress.jsonl`).
- `74 passed` across focused sandbox runtime/example/hostile-input tests.
- `1370 passed, 24 skipped, 840 subtests passed` across `tests/` after installing the viewer's locked dependencies.
- Live Docker probe: the target successfully created a forged lookalike under `/sandbox/output`, while the host ledger remained unchanged and its directory was absent from target mounts.
- The probe left no new sandbox containers or networks.

## Risk

Low and scoped to the stock container runtime's temporary-directory layout. Both directories remain owned by the same per-case temporary session and are removed together during session cleanup.

### DIFF
--- a/assert_ai/integrations/sandbox/runtime.py
+++ b/assert_ai/integrations/sandbox/runtime.py
@@ -11,7 +11,8 @@ complementary process and network boundary for everything else:
 * a private Docker network with no gateway, so neither the host nor public
   internet is directly reachable;
 * a deny-by-default HTTP(S) proxy that records every proxy-aware egress attempt;
-* policy and mock files mounted read-only, with a separate writable output mount;
+* policy and mock files mounted read-only, with a separate writable target-output
+  mount and a host-only network-evidence ledger;
 * optional host-side model proxy, so a real provider credential never enters the
   container.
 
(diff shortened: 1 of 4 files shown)


changliu2 (approved):
Reviewed exact head `dc64d90` against current `main`. The host-owned egress ledger remains outside every target mount, the merge is clean, and the targeted sandbox tests pass. No blockers found.

PR #327, opened 2026-08-20
fix(sandbox): correlate mediation by test case

jakepresent:
## Problem

ASSERT carried a `case_id`, but the sandbox path did not propagate and enforce one consistent identity end to end. That made case-bound mocks incomplete, allowed ambiguous rule selection, and left evidence unable to prove which case rule fired.

## Fix

- propagate the effective ASSERT case ID through `SandboxedEndpointSession`, the stock container environment, and the optional JSON request sent to an already-running sandbox endpoint;
- preserve the ordinary `target.endpoint` request shape when no case ID is configured;
- use one case-ID resolver for mock selection and evidence, and reject conflicting legacy/session IDs before any pass-through tool executes;
- define rule precedence as exact case ID, matching case glob, generic fallback, argument specificity, then file order;
- partition built-in scenario state by case while preserving the legacy `current_state(scenario)` extension contract;
- serialize matching and state transition so parallel calls cannot consume the same scenario step;
- include `mock_source`, replay detail, and `matched_case_id` in judge-visible evidence; and
- keep the packaged and example stock servers aligned.

## Verification

- Focused case, endpoint, sandbox, and runtime suites: `124 passed`.
- Full exact-head Python suite: `1461 passed, 23 skipped, 840 subtests passed`.
- Regression coverage includes endpoint payload propagation, exact/glob/generic precedence, conflicting identities, legacy backend transitions, parallel scenario resolution, and judge-visible matched-rule evidence.
- Branch includes current `main`, including the merged host-owned egress ledger correction.

## Risk

The endpoint `case_id` field is optional. Existing ordinary endpoints and mock rules without a case binding retain their previous behavior. Legacy custom scenario backends remain usable, but cannot gain per-case state isolation until they adopt the new case-aware contract.

A clean local stock-image rebuild remains blocked by the local Docker builder's external PyPI TLS failure. The final combined correction tree was separately exercised with the exact branch source overlaid onto the existing stock image; clean-image CI is owned by #328.

### DIFF
--- a/assert_ai/core/session.py
+++ b/assert_ai/core/session.py
@@ -677,7 +677,9 @@ class CallableSession:
 class HTTPEndpointSession:
     """Invokes an HTTP endpoint as the eval target.
 
-    POST {"message": text, "history": [...]} to the URL.
+    POST {"message": text, "history": [...]} to the URL. Sandboxed callers may
+    also include an optional ``case_id`` so a reusable endpoint can select and
+    report case-specific behavior.
     Expects {"response": "..."} back. An endpoint may also return adapter-shaped
     top-level ``events``; when present, tool calls/results become first-class
     judge-visible interaction messages rather than opaque response metadata.
@@ -692,6 +694,7 @@ class HTTPEndpointSession:
         system_prompt: str | None = None,
         message_timeout_s: float | None = None,
         allow_private: bool = False,
+        case_id: str | None = None,
     ) -> None:
         from assert_ai.core.security import validate_endpoint_url
 
@@ -700,6 +703,7 @@ class HTTPEndpointSession:
         self._headers = headers or {}
         self._system_prompt = system_prompt
         self._timeout_s = message_timeout_s
+        self._case_id = case_id
         self._session = None  # aiohttp.ClientSession
         self._resolver = None
 
@@ -757,6 +761,8 @@ class HTTPEndpointSession:
         ]
 
         payload = {"message": user_text, "history": history}
+        if self._case_id:
+            payload["case_id"] = self._case_id
 
         try:
             async with self._session.post(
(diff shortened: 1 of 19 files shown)


changliu2 (changes requested):
Two blockers on exact head `46d583a`:

1. Endpoint sandbox targets do not receive the per-test-case ID. The endpoint session still posts only `message` and `history`, so case-bound mocks cannot vary correctly across an already-running endpoint.
2. `ScenarioBackend.current_state()` now accepts an optional case ID, but callers always pass it; legacy subclasses that override the prior one-argument method raise `TypeError`.

Please also align wildcard/no-case matching and evidence case-ID precedence so mock selection and recorded evidence cannot disagree.

changliu2 (changes requested):
The endpoint propagation, legacy `ScenarioBackend.current_state()` compatibility, conflict handling, concurrency isolation, and evidence case-proof fixes now look resolved. One blocker remains: a mock rule with `case_id: "*"` still matches a call with no case ID because the missing ID is matched as an empty string and `*` succeeds. This contradicts the CLI/README guarantee that case-bound rules cannot match uncorrelated calls and can select the wrong mock. Please explicitly skip every case-bound rule when `call.case_id` is absent, and add library and CLI coverage for `case_id: "*"`.

jakepresent:
Fixed the remaining wildcard case-binding issue in `5781f1e`: every case-bound rule, including `case_id: "*"`, now requires a non-empty correlated case ID. I added library and CLI regressions for missing and empty IDs, and all checks are green. @changliu2 ready for re-review.

changliu2 (approved):
Re-reviewed exact head `5781f1e`. The wildcard case-binding blocker is resolved: every case-bound rule, including `case_id: "*"`, is now rejected when the call has no case ID, and the library/CLI coverage exercises both uncorrelated and correlated wildcard paths. The earlier endpoint propagation, legacy backend compatibility, conflict handling, concurrency isolation, and evidence case-proof fixes remain intact. Approved.

PR #328, opened 2026-08-20
ci(sandbox): run real Docker containment tests

jakepresent:
## Problem

ASSERT's real Docker sandbox tests were opt-in, and the ordinary unit workflow could collect and skip all of them while still reporting success. A path-filtered workflow also could not serve as a reliable required check.

## Fix

Add an always-present GitHub Actions workflow that:

- runs on every pull request and every push to `main`, avoiding path-filter omissions and pending required checks;
- explicitly fails if the Docker daemon is unavailable;
- makes the test module fail instead of skip when `ASSERT_RUN_DOCKER_TESTS=1` but Docker cannot run;
- executes every `tests/test_sandbox_*docker*.py` module rather than one hard-coded file;
- uploads a JUnit receipt for the exact tests that ran; and
- always checks for leftover target/relay containers and networks, including after failure.

## Verification

- With no opt-in flag, the four live tests skip during ordinary local test runs.
- With the flag enabled and a deliberately failing fake Docker command, collection fails instead of passing with skips.
- Exact-head GitHub Actions ran `Sandbox: real Docker containment` successfully.
- The final combined #327/#328/#329 tree ran all four real-Docker tests with no failures or skips and left no sandbox containers or networks.
- Branch includes current `main`, including the merged host-owned egress ledger correction.

## Remaining repository gate

The current `main` ruleset requires review but does not require status checks. After this workflow merges, `Sandbox: real Docker containment` should be added as a required status check with strict base-branch freshness. The workflow now runs on every PR, so making it required will not leave unrelated PRs waiting for a path-filtered check that never starts.

A clean local image rebuild remains blocked by the local Docker builder's external PyPI TLS failure; the clean build and test path passed in GitHub Actions.

### DIFF
--- a/.github/workflows/sandbox-docker.yml
+++ b/.github/workflows/sandbox-docker.yml
@@ -14,7 +14,9 @@ permissions:
   contents: read
 
 concurrency:
-  group: sandbox-docker-${{ github.head_ref || github.ref }}
+  # Branch names are not unique across forks. Scope PR runs to the PR number so
+  # one contributor cannot cancel another contributor's required Docker gate.
+  group: sandbox-docker-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/sandbox-docker.yml
+++ b/.github/workflows/sandbox-docker.yml
@@ -38,7 +38,23 @@ jobs:
         run: docker info
 
       - name: Run real Docker sandbox tests
-        run: pytest tests/test_sandbox_*docker*.py -x -q --junitxml=sandbox-docker-results.xml
+        run: pytest tests/test_sandbox_*docker*.py --runxfail -x -q --junitxml=sandbox-docker-results.xml
+
+      - name: Verify Docker test receipt
+        if: always()
+        shell: python
+        run: |
+          import xml.etree.ElementTree as ET
+
+          root = ET.parse("sandbox-docker-results.xml").getroot()
+          suites = [root] if root.tag == "testsuite" else list(root.findall(".//testsuite"))
+          totals = {
+              key: sum(int(suite.attrib.get(key, 0)) for suite in suites)
+              for key in ("tests", "failures", "errors", "skipped")
+          }
+          print(totals)
+          if totals["tests"] <= 0 or any(totals[key] for key in ("failures", "errors", "skipped")):
+              raise SystemExit("real Docker receipt must contain executed tests with no failures or skips")
 
       - name: Preserve Docker test receipt
         if: always()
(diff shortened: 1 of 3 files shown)


changliu2 (changes requested):
Two CI blockers on exact head `0db98e0`:

1. When Docker is unavailable, `tests/test_sandbox_runtime_docker.py` skips all four tests and the workflow still succeeds. Add an explicit failing Docker preflight so an outage cannot produce a false green.
2. The workflow path filters omit direct sandbox dependencies such as `assert_ai/core/security.py`. Prefer `assert_ai/**` or include the complete dependency surface.

Until these are fixed, green CI does not reliably prove real Docker containment.

changliu2 (changes requested):
The two previous blockers are resolved: Docker unavailability now fails explicitly, xfail/skip cannot mask the proof, the receipt confirms four executed tests, and removing path filters covers all relevant changes. One required-check reliability blocker remains: the concurrency group uses only `github.head_ref`, so PRs from different forks with the same branch name can cancel each other’s required check. Please key PR runs by `github.event.pull_request.number`, with `github.ref` as the non-PR fallback.

jakepresent:
Fixed the concurrency collision in `3fa87f9`: PR runs now key the Docker job by pull-request number, with `github.ref` as the push and manual-run fallback. I added a workflow regression, and all checks including real Docker containment are green. @changliu2 ready for re-review.

changliu2 (changes requested):
The PR-number concurrency key fixes the cross-fork cancellation blocker, and the earlier Docker preflight, path coverage, xfail/skip receipt, and cleanup fixes remain sound. One new false-green path was introduced on this head: `tests/test_sandbox_docker_workflow.py` matches the workflow selector `tests/test_sandbox_*docker*.py`. Because the receipt only requires a positive test count, deleting or renaming all real container tests would still run this static YAML test and let the "real Docker containment" gate pass with one test. Please keep the static workflow test outside that selector (for example, rename it) or explicitly select/mark the real Docker containment tests, so the receipt cannot be satisfied without executing one.

jakepresent:
Fixed the false-green selector on exact head `e80bdef`: the Docker workflow now explicitly runs only `tests/test_sandbox_runtime_docker.py`, and the static workflow regression locks that exact selector. The GitHub receipt contains the four real container tests with zero failures, errors, or skips, and all checks are green. @changliu2 ready for re-review.

changliu2 (approved):
Re-reviewed exact head `e80bdef`. The static-test false-green blocker is resolved: the workflow now selects only `tests/test_sandbox_runtime_docker.py`, the regression test pins that selector, and the hosted receipt contains four executed tests with no failures, errors, or skips. The earlier Docker-daemon preflight, all-change path coverage, xfail/skip protection, cleanup proof, and PR-scoped concurrency fixes remain sound. Approved.

PR #329, opened 2026-08-20
fix(sandbox): make tool decisions host-authoritative

jakepresent:
## Problem

The stock sandbox trusted the evaluated target to choose pass/mock/block and return its own action evidence. Read-only policy files did not make that evidence authoritative: target code could omit or forge endpoint events, and the writable target ledger was not an independent record.

## Approach

Add an explicit `target.host_action_mediation: true` path for stock containers:

- run policy evaluation, mock resolution, and the action ledger in an authenticated service on the ASSERT host;
- route the target's `AgentHooksToolHost` through `RemoteActionMediator` over the existing narrow relay;
- require registration of the ASSERT-owned case ID during startup;
- append the attempted call and pass/mock/block decision before returning to the target;
- keep the ledger outside every target-visible mount;
- replace target-supplied tool events with host ledger rows;
- preserve completed and incomplete action rows when a turn fails; and
- fail startup or the turn rather than silently accepting an opted-in target that did not use the host mediator.

The evidence contract remains bounded:

- attempts and pass/mock/block decisions are host-authoritative;
- mock/block results are host-generated and authoritative;
- passed tools execute against disposable state inside the target, so their results remain explicitly `target_reported` and `result_authoritative: false`; and
- arbitrary in-process calls that bypass the mediator and are omitted by the target are not claimed as observed. Container isolation, deny-by-default egress, and host-only real credentials remain the containment backstop.

## Review hardening

Adversarial review found and this head fixes:

- target-forged or conflicting case IDs;
- duplicate call IDs consuming scenario state;
- completed and incomplete rows draining out of attempt order;
- host evidence being lost when fail-closed validation raised;
- an opted-in image starting without a host-mediator client;
- an untracked pass executor running behind an incorrect execution claim;
- a fail-closed gate coupled to rendered evidence wording;
- an unrelated host row hiding an unmatched target action;
- the same call ID being reused for different tool names or arguments;
- target actions with missing IDs;
- target-controlled tool evidence being accepted without per-action host reconciliation;
- target startup racing the new host-mediation relay on native Linux Docker;
- whole-row sanitization changing structural call IDs and tool names;
- host evidence replacing all actions at the end of a multi-action turn and changing chronology;
- unbounded host worker threads and stalled target connections;
- large or unserializable passed-tool results preventing completion evidence;
- policy failures occurring before an attempted action was recorded;
- omitted tool-result arguments conflicting with the matching tool call;
- falsey non-object arguments being silently rewritten to `{}`;
- pending completion evidence being removed before its durable transition succeeded;
- host/target argument comparison collapsing distinct credential-shaped values after redaction;
- repeated target tool-call occurrences collapsing into one host-ledger claim;
- generated public IDs colliding with target-chosen IDs;
- credential-shaped argument keys bypassing value-only sanitization;
- non-finite JSON arguments failing evidence generation after a pass decision;
- the target reordering otherwise valid host action claims;
- cumulative action volume exhausting host ledger memory or disk;
- slow-trickle connections defeating inactivity-only timeouts;
- unbounded sandbox endpoint responses being buffered on the host; and
- per-turn public action IDs restarting each turn, so distinct actions in a multi-turn transcript shared one ID.

Host-mode reconciliation now captures ordered canonical digests before sanitization, requires each target tool-call occurrence to match one private host claim in host-sequence order, and rejects duplicate calls even when their contents are identical. The digests never enter dataclass serialization, logs, or artifacts. Every persisted action uses a host-owned sequence alias, and credential-shaped values and mapping keys are sanitized. Canonical identity is validated before mediation, each case is limited to 64 actions, mediator connections have an absolute deadline, and sandbox endpoint responses are capped at 4 MiB. Results may inherit arguments from their matching call, matching target events are replaced in place, and host-only attempts remain visible before the final response.

## Verification

- Focused sandbox, host-mediator, endpoint, and hostile-input suites: `180 passed`.
- Focused final hardening regressions: `8 passed`.
- Full exact-head Python suite: `1512 passed, 24 skipped, 840 subtests passed`.
- The exact-head clean-image GitHub workflow passed all five real Docker tests with zero failures, errors, or skips, verified the JUnit receipt, and removed all sandbox resources.
- A new multi-turn Docker proof drives two turns through one live sandbox session and asserts host-mediated actions stay individually addressable (`host-action-0` through `host-action-3`), that host ledger sequence keeps advancing across turns, and that the irreversible action stays mocked on every turn. It runs in the Docker gate, not just locally.
- The Docker proof verified host decisions for pass/mock actions, target-reported pass results, host-generated mock results, host-only ledgers, egress containment, credential isolation, and cleanup.
- Build, CodeQL, Tier 1, and the Python 3.11-3.13 install matrix passed on Linux, macOS, and Windows.
- No sandbox containers or networks remained.

A clean local stock-image rebuild remains blocked by the local Docker builder's external PyPI TLS failure. The clean-image proof above comes from the GitHub-hosted runner against this exact head.

## Landing

This PR is now based on `main`, including the merged #327 case-correlation and #328 mandatory Docker-gate changes. Its retargeted diff contains only the host-authoritative action-mediation change.

## Deliberate follow-up

This PR deliberately remains an interim `RemoteActionMediator` compatibility path. The stronger follow-up architecture is a mandatory host-owned MCP gateway that owns tool discovery, policy, mocks, credentials, execution, and evidence. That would make passed execution independently authoritative, but changes the customer onboarding and tool-packaging contract and should be reviewed separately.


changliu2 (changes requested):
The fail-closed gate on exact head `a412143` accepts the presence of any host-ledger row, then drops the target tool events. An unmatched target action can therefore disappear when the host ledger contains only an unrelated action. Correlate every target action/tool-call ID against the trusted host-ledger IDs and fail closed on any unmatched action.

This PR is also stacked on #327, so please rebase and revalidate after the case-correlation blockers there are fixed. The exact head currently has only Tier 1 CI.

liamcrumm:
I wanted to bring up with you, I think having something like RemoteActionMediator is better than nothing, but I think we should likely model this as a separate MCP tool gateway. RemoteActionMediator is moving the policy decision to the host, but the passed tools still execute inside the agents container and their results are target-reported, so we're depending on the agent to execute them, report the result, and not bypass our wrapper. We could have a separate docker container which acts as an mcp tool gateway. The caller just needs to configure this gateway as an mcp server for their agent. It will provide MCP tools/list and tools/call methods, so all tools route through this separate container. Everything is blocked other than the model proxy and this mcp tool gateway. It would reduce the integration code, since instead of relying on the integrator to wrap every tool call, it would only need to configure the MCP server. You could move tool implementations, credentials, policy enforcement, mocks, execution, and the ledger itself into this gateway. 

Besides this main point, I think policy should be context aware, so you can decide pass/mock/block using tool name, arguments, ASSERT case ID, and scenario state instead of mainly just tool name. You should also restrict egress by both host and port, so that allow-listing one hostname does not permit every service running on it.

jakepresent:
Updated and retargeted #329 to current `main` on exact head `d14a231`. The first hosted Docker run exposed a target-startup race with the new host-mediation relay; `d14a231` now retries transport-only registration failures, and the final clean-image run executed all four real Docker tests with zero failures, errors, or skips and clean teardown. The full local suite passed `1487 passed, 23 skipped, 840 subtests passed`. @changliu2 @liamcrumm ready for final exact-head review.

liamcrumm:
I mentioned something similar above, but wanted to clarify in this review what I see as the long-term direction. `RemoteActionMediator` is a reasonable temporary compatibility path, but I do not think it should become the primary sandbox architecture. The final design should move the tool boundary outside the agent into a mandatory host-owned MCP gateway that owns tool discovery, policy, mocks, credentials, execution, and evidence. This would remove the need for ASSERT-specific wrappers inside the agent and make passed execution authoritative rather than target-reported. For this review, I will assess the current implementation as an interim solution rather than the final architecture.

assert_ai/integrations/sandbox/host_mediator.py
```diff
@@ -0,0 +1,268 @@
...
+            raise ValueError(
+                "mediated call case_id does not match the ASSERT-owned sandbox case"
+            )
+        with self._lock:
+            if call_id in self._seen:
+                raise ValueError(f"duplicate mediation call id {call_id!r}")
+            decision = self._mediator.plan(pre_context)
+            sequence = self._next_sequence
+            self._next_sequence += 1
+            row = sanitize_untrusted_value({
```
liamcrumm: Could we avoid sanitizing the entire ledger row? Distinct credential-shaped `id` values can both become `[REDACTED]`, causing reconciliation to treat them as duplicates after the rows have been drained, so the actions never reach the transcript. Structural fields such as `id` and `tool` should remain unchanged, with sanitization limited to payload fields such as `args` and `returned`.

assert_ai/integrations/sandbox/session.py
```diff
@@ -184,17 +261,76 @@ async def run_turn(self, messages: list[Message]) -> TurnResult:
...
+                if mismatched_ids:
+                    details.append(
+                        "host/target tool or argument mismatch for: "
+                        + ", ".join(mismatched_ids)
+                    )
+                raise RuntimeError(
+                    "host action mediation is enabled, but the target returned tool events "
+                    "not accounted for by the host mediator (" + "; ".join(details) + ")"
+                )
+            # The evaluated target controls endpoint events. In host-mediated
```
liamcrumm: Could we replace each target tool event in place rather than inserting all host rows before the final response? In a multi-action turn, this moves intermediate assistant text ahead of actions it originally followed, giving the judge the wrong chronology. Matching and replacing events by call ID would preserve the original order.

assert_ai/integrations/sandbox/host_mediator.py
```diff
@@ -0,0 +1,268 @@
...
+        mediator,
+        ledger_path=ledger_path,
+        expected_case_id=case_id,
+    )
+    handler = type(
+        "SandboxHostMediatorHandler",
+        (_HostMediatorHandler,),
+        {"ledger": ledger, "access_token": access_token},
+    )
+    server = ThreadingHTTPServer(("0.0.0.0", 0), handler)
```
liamcrumm: Could we bound connections and worker threads here and set a per-connection timeout? `ThreadingHTTPServer` creates an unbounded host thread for each stalled connection from the target, allowing it to exhaust host resources outside the container limits.

assert_ai/integrations/sandbox/host_mediator.py
```diff
@@ -0,0 +1,268 @@
...
+            self._send(404, {"error": "not_found"})
+
+    def do_POST(self) -> None:
+        supplied = self.headers.get("authorization", "")
+        if not secrets.compare_digest(supplied, f"Bearer {self.access_token}"):
+            self._send(401, {"error": "invalid_mediator_token"})
+            return
+        try:
+            length = int(self.headers.get("content-length", "0") or 0)
+            if length <= 0 or length > _MAX_REQUEST_BYTES:
```
liamcrumm, line 496: Could we handle the completion size before executing a passed tool? A tool can execute successfully and then have `/complete` rejected because its result exceeds this limit, leaving the ledger marked missing and execution unknown after a real side effect.

assert_ai/integrations/sandbox/host_mediator.py
```diff
@@ -0,0 +1,268 @@
...
+        session = pre_context.get("session") or {}
+        claimed_case_id = str(session.get("case_id") or "") or None
+        if self.expected_case_id is not None and claimed_case_id != self.expected_case_id:
+            raise ValueError(
+                "mediated call case_id does not match the ASSERT-owned sandbox case"
+            )
+        with self._lock:
+            if call_id in self._seen:
+                raise ValueError(f"duplicate mediation call id {call_id!r}")
+            decision = self._mediator.plan(pre_context)
```
liamcrumm: Could we record the attempt before evaluating policy? A matcher error caused by the submitted arguments returns 400 before any ledger row is written, so the attempted action disappears from the authoritative record.

assert_ai/integrations/sandbox/session.py
```diff
@@ -52,6 +53,77 @@ def _validate_target_action_case(
...
+                    raise RuntimeError("target tool call must be an object")
+                _record_action_claim(
+                    claims,
+                    call_id=call.get("id"),
+                    tool=call.get("function"),
+                    args=call.get("arguments"),
+                    source="target-reported",
+                )
+        if message.get("role") == "tool":
+            _record_action_claim(
```
liamcrumm: Tool results that omit arguments should inherit them from the matching tool call. `tool_args` is optional, but the serializer supplies `{}`, which reconciliation treats as conflicting with the call’s actual arguments and rejects an otherwise valid response.

assert_ai/integrations/sandbox/host_mediator.py
```diff
@@ -0,0 +1,268 @@
...
+                "mediator registration case_id does not match the ASSERT-owned sandbox case"
+            )
+        with self._lock:
+            self._registered = True
+
+    def mediate(self, pre_context: dict[str, Any]) -> MediationDecision:
+        tool_call = pre_context.get("tool_call") or {}
+        call_id = str(tool_call.get("id") or "").strip()
+        tool = str(tool_call.get("name") or "").strip()
+        args = tool_call.get("args") or {}
```
liamcrumm: The raw argument value should be validated before applying a default. `tool_call.get("args") or {}` rewrites falsey non-object values such as `[]` to `{}`, so the authoritative ledger can record different arguments from those submitted.

assert_ai/integrations/sandbox/host_mediator.py
```diff
@@ -0,0 +1,268 @@
...
+    def complete(
+        self,
+        *,
+        call_id: str,
+        returned: Any,
+        is_error: bool,
+        real_executed: bool,
+    ) -> None:
+        with self._lock:
+            row = self._pending.pop(call_id, None)
```
liamcrumm: The pending row should not be removed until the completion transition has been written successfully. If `_append_transition()` fails after this `pop`, the evidence for an already-executed action is lost and a retry is rejected as an unknown call ID.

jakepresent:
Addressed Liam's eight exact-head comments in `4e532bc`:

- structural call IDs and tool names now survive sanitization while payload fields remain redacted;
- target action evidence is replaced in place, result events inherit omitted arguments from their matching calls, and falsey non-object arguments are rejected without rewriting;
- every valid action attempt is persisted before policy evaluation, policy failures produce a fail-closed ledger row, and completion state is removed only after the durable transition succeeds;
- large or unserializable passed-tool results use a bounded completion envelope with size/hash evidence; and
- the host mediator caps worker threads and times out stalled target connections.

The implementation remains explicitly scoped as the interim `RemoteActionMediator` compatibility path. The mandatory host-owned MCP gateway, context-aware policy, and host-plus-port egress remain separate follow-ups.

Verification on exact head: `1500 passed, 23 skipped, 840 subtests passed`; focused sandbox suites passed `167`; all GitHub checks are green; and the clean-image Docker job executed 4 tests with zero failures, errors, or skips and clean teardown. @liamcrumm @changliu2 ready for exact-head re-review.

changliu2 (changes requested):
On exact head `4e532bc`, the original credential-shaped ID collision, action chronology reordering, and omitted tool-result argument failures are fixed. Two reconciliation gaps remain:

1. Host and target arguments are still sanitized before comparison, so distinct credential-shaped argument values collapse to `[REDACTED]` and an unmatched action can be accepted.
2. Repeated target tool calls with the same ID, tool, and arguments collapse to one host-ledger row instead of failing one-to-one occurrence reconciliation.

Please reconcile raw non-persisted claims or canonical digests for IDs, tools, and arguments; enforce one-to-one tool-call occurrence matching; and sanitize only the evidence persisted to logs and artifacts.

jakepresent:
Fixed both remaining reconciliation gaps on exact head `88fd1de`. Target and host action identities now use ordered, pre-redaction canonical digests that never enter logs or artifacts; duplicate target tool-call occurrences fail one-to-one reconciliation, and persisted credential-shaped IDs use safe host sequence aliases. Exact-head verification is green: `1504 passed, 23 skipped, 840 subtests passed`, focused sandbox coverage passed `172`, and the clean-image Docker job executed 4 tests with zero failures, errors, or skips and clean teardown. @changliu2 @liamcrumm ready for exact-head re-review.

jakepresent:
Closed the remaining clean-sheet audit findings on exact head `6b73597`: all persisted action IDs are now host-owned aliases, credential-shaped mapping keys are sanitized, canonical arguments fail before mediation, target action order must follow host sequence, and host action volume is bounded. Mediator connections now have an absolute deadline that defeats slow-trickle clients, and sandbox endpoint responses are capped before JSON buffering. Exact-head verification is green with `1511 passed, 23 skipped, 840 subtests passed`, 179 focused sandbox tests, and 4 clean-image Docker tests with zero failures, errors, or skips and clean teardown. @changliu2 @liamcrumm ready for exact-head re-review.

jakepresent:
One more fix on top of the last round, at exact head `af1a298`. Public endpoint action IDs restarted at zero on every turn, so in a multi-turn transcript a benign early action and a sensitive later action both appeared as `endpoint-action-0` and the judge could not tell them apart. Public IDs are now unique for the life of the session. This was a defect in my own previous fix, found by probing the new code rather than by re-reading it.

Exact-head verification: `1512 passed, 23 skipped, 840 subtests passed`, 180 focused sandbox tests, and 4 clean-image Docker tests with zero failures, errors, or skips and clean teardown. @changliu2 @liamcrumm ready for exact-head re-review.

jakepresent:
Closed the verification gap I flagged earlier. Every previous host-mediation proof was single-turn, so cross-turn behavior was never exercised against a real container. `tests/test_sandbox_multiturn_docker.py` now drives two turns through one live Docker sandbox and asserts that host-mediated actions stay individually addressable across the session (`host-action-0` through `host-action-3`), that the host ledger sequence keeps advancing rather than restarting per turn, and that the irreversible `send_message` stays mocked on both turns. It runs inside the Docker gate rather than only locally, so the receipt now covers five real-container tests instead of four.

I also confirmed the earlier cross-turn ID fix by negative control: reverting it makes the new unit regression fail with both turns reporting `endpoint-action-0`, and restoring it passes.

Exact head `b6af2f6`: `1512 passed, 24 skipped, 840 subtests passed`, and the Docker gate reports `{'tests': 5, 'failures': 0, 'errors': 0, 'skipped': 0}` with clean teardown. @changliu2 @liamcrumm ready for exact-head re-review.

changliu2 (changes requested):
Re-reviewed exact head `b6af2f6`. The prior sanitized-argument collision, duplicate-call occurrence, chronology, omitted-result-argument, and cross-turn ID/sequence issues are fixed, and the five-test real-Docker proof is meaningful.

One supported-path blocker remains: the documented `AgentHooksToolHost` framework/HTTP-adapter boundary accepts arbitrary Python values in `dict[str, Any]`. `RemoteActionMediator` silently stringifies unsupported values for host policy evaluation, then executes the original object. The host can therefore approve and record arguments different from those actually used by the tool.

Please canonicalize and validate arguments once as JSON-native values at the public boundary, then use that exact object for mediation, execution, completion, and evidence.

jakepresent:
Addressed the remaining argument-boundary blocker on exact head `8f50c81`. `AgentHooksToolHost.call_tool` now creates one strict JSON-native snapshot before building the mediation context (`allow_nan=False`), so unsupported values fail before mediation or execution and the host policy, real tool, and recorded evidence all use the same canonical arguments; the new regression exercises this through the actual `RemoteActionMediator` and host ledger. Exact-head verification is green: the full local suite passed `1514 passed, 24 skipped, 840 subtests passed`, and all GitHub checks pass, including the real-Docker containment gate. @changliu2 ready for exact-head re-review.

assert_ai/integrations/sandbox/host_mediator.py
liamcrumm: The size issue is fixed, but completion results still need strict JSON validation. A passed tool returning `NaN` or `Infinity` writes non-standard JSON into the host ledger and inference artifact, which strict JSON consumers reject. These values should be rejected or safely encoded before persistence.

assert_ai/integrations/sandbox/host_mediator.py
```diff
@@ -0,0 +1,459 @@
...
+            for row in incomplete:
+                self._append_transition({"phase": "completion_missing", **row})
+            rows = sorted(
+                [*self._ready, *incomplete],
+                key=lambda row: int(row["sequence"]),
+            )
+            claims = [self._claims[str(row["id"])] for row in rows]
+            public_rows = [_public_ledger_row(row) for row in rows]
+            self._ready.clear()
+            self._pending.clear()
```
liamcrumm, line 359: Pending pass calls should remain completable after the failure-path drain. Clearing `_pending` here causes a tool that finishes during shutdown to have `/complete` rejected and remain recorded as `unknown` even though it executed. Keep pending entries until the target stops, then perform a final drain.

assert_ai/integrations/sandbox/host_mediator.py
```diff
@@ -0,0 +1,459 @@
...
+        session = pre_context.get("session") or {}
+        claimed_case_id = str(session.get("case_id") or "") or None
+        if self.expected_case_id is not None and claimed_case_id != self.expected_case_id:
+            raise ValueError(
+                "mediated call case_id does not match the ASSERT-owned sandbox case"
+            )
+        with self._lock:
+            if call_id in self._seen:
+                raise ValueError(f"duplicate mediation call id {call_id!r}")
+            if self._next_sequence >= self.max_actions:
```
liamcrumm, line 140: The action limit is checked before the attempt is recorded. After 64 actions, later requests are rejected with no ledger row, leaving the authoritative record incomplete. Could we record the rejected attempt before returning the limit error?

assert_ai/integrations/sandbox/session.py
```diff
@@ -52,6 +56,276 @@ def _validate_target_action_case(
...
+        raise RuntimeError("private action claims contain events missing from the transcript")
+
+    remaining: list[dict[str, Any]] = []
+    for call_id in host_order:
+        for kind in ("call", "result"):
+            host_message = host_messages[call_id].get(kind)
+            if host_message is not None and (call_id, kind) not in inserted:
+                remaining.append(host_message)
+    remaining.extend(passthrough)
+    if not remaining:
```
liamcrumm, line 340: The in-place replacement works when both events are present, but an omitted tool-result event is appended near the final response. This can still place assistant prose before the result it describes. Could we insert the host result immediately after the matching call when the target omits it?

assert_ai/integrations/sandbox/mediation_setup.py
```diff
@@ -169,6 +170,9 @@ def _load_target(data: Any) -> TargetSpec:
     model_proxy = data.get("model_proxy")
     if model_proxy is not None and not isinstance(model_proxy, Mapping):
         raise SetupError("`target.model_proxy` must be a mapping")
+    host_action_mediation = data.get("host_action_mediation", False)
```
liamcrumm, line 177: Could we reject `host_action_mediation` for endpoint targets instead of ignoring it?

assert_ai/integrations/sandbox/session.py
```diff
@@ -52,6 +56,276 @@ def _validate_target_action_case(
...
+        target_claim = target_actions[call_id]
+        host_claim, _public_id, host_position = host_match
+        if (
+            target_claim.tool_digest != host_claim.tool_digest
+            or target_claim.arguments_digest != host_claim.arguments_digest
+        ):
+            mismatch_positions.append(str(position))
+        if host_position <= last_host_position:
+            reordered_positions.append(str(position))
+        last_host_position = host_position
```
liamcrumm, line 210: Parallel tool calls can complete in a different order from when they were requested. Reconciliation currently rejects this as a sequence mismatch even though both calls were mediated correctly. Could we match each completed action to its host record without requiring completion order to match request order?

changliu2 (dismissed):
Re-reviewed exact head `b3c66dad`. The prior canonical-argument blocker is fixed, and five of the six follow-up reconciliation/lifecycle issues are addressed.

One supported-path blocker remains in `assert_ai/integrations/sandbox/session.py:160-202`: result-only target action claims are added to `target_action_order`, so valid parallel calls requested as A then B but completed and reported as result B then result A are rejected as diverging from host request order. The new regression covers `[call A, call B, result B, result A]`, but not the supported result-only form `[result B, result A]` with supplied arguments.

I reproduced this at the current head: `_reconcile_action_claims([result_B, result_A], [host_A, host_B], ...)` raises `target action order diverges from host sequence at occurrences: 2` even though both results match one host-mediated call exactly.

Please enforce host sequence ordering only for explicit target call occurrences. Standalone result occurrences should reconcile one-to-one by call ID, tool, and canonical arguments without treating completion order as request order.

liamcrumm (approved):
Most of the previous findings are now addressed, and the added regressions and Docker coverage are meaningful. Two issues remain to follow up on: valid parallel tool calls are rejected when they complete in a different order from their request order, and completion handling can still record an executed action as `missing/unknown` when the completion envelope exceeds the request limit. The longer-term direction should remain a mandatory host-owned MCP tool gateway rather than treating `RemoteActionMediator` as the final architecture.

liamcrumm (approved):
Approved. The previous review findings have been addressed, including parallel completion ordering, host evidence handling, lifecycle finalization, and endpoint configuration validation. The remaining custom-result serialization edge can be handled as follow-up work.

changliu2 (approved):
approving

PR #336, opened 2026-08-26
refactor(deps): isolate optional dependency environments

tangym:
## Summary

Updates vulnerable frontend and Python dependencies, then narrows published ASSERT dependency metadata to runtime product features. Framework examples and optional scripts now own their dependencies through adjacent requirements files instead of one combined installation environment.

## Motivation / linked issue

Reduce dependency vulnerability exposure, remove cross-framework resolver coupling, and make optional dependency ownership explicit.

## Changes

- Update viewer dependencies and pin patched SvelteKit and PostCSS releases.
- Replace the website's vulnerable Keyv cache chain through `file-entry-cache`.
- Remove framework-only `langgraph`, `dspy`, and `examples` extras from published ASSERT metadata.
- Remove the unused `regression` extra and script-only packages from base and `analysis`.
- Split heavyweight local embedding support into `assert-ai[embeddings]`.
- Cap ACS Generator below 0.4 because ASSERT uses the 0.3 `GenerationEngine` API.
- Keep `all` limited to product features: tracing, analysis, embeddings, Azure identity, and ACS.
- Add independently installable requirements files for the flagship LangGraph planner, Azure Doc QA, Bank Manager, the default OpenAI auto-trace path, Foundry hosting, and optional scripts.
- Update canonical setup, CI, dev container, assistant guides, and example documentation to use root product extras plus adjacent requirements files.

The multi-framework auto-trace gallery keeps its alternate framework modules as explicit opt-ins: install the framework and matching OpenInference instrumentor named in the gallery table. CrewAI is not committed to a requirements manifest while its ChromaDB dependency has no patched release.

## Testing

- Root and isolated requirements audits: 0 known vulnerabilities
- Python: 1,453 passed, 22 skipped, 1 Docker-host test deselected
- Every requirements file resolves and imports on Python 3.11
- `uv lock --check`
- Viewer: `npm audit`, `npm ci`, `npm run check`, `npm run build`
- Website: `npm audit`, `npm ci`, `npm run build`
- Package: `python -m build`, `twine check dist/*`

## Checklist

- [x] Tests pass locally.
- [x] Setup and example documentation updated for the new dependency boundaries.
- [x] No breaking runtime API change.


jakepresent (changes requested):
Two blockers on exact head `ed539e3`:

1. **The documented Bank Manager install leaves an incompatible environment.** I ran the README sequence in a clean Python 3.11 venv:

   ```bash
   python -m pip install -e ".[acs,otel]"
   python -m pip install -r examples/bank_manager_agent_control/requirements.txt
   python -m pip check
   ```

   Both install commands exit successfully, but `pip check` fails because `langchain-azure-ai>=1.2.4` pulls `azure-ai-projects 2.5.0`, which requires `openai>=3.0.0`, while the installed `litellm 1.98.0` requires `openai>=2.20.0,<3.0.0`. The resulting environment has `openai 3.5.0` and is inconsistent. This contradicts the PR's requirements-resolution claim and can fail after an apparently successful setup. Please use a compatible version set or isolate that provider path, and add a required clean-venv test of the documented two-step install followed by `pip check`.

2. **The package silently removes existing public extras without a compatibility or release migration.** The built wheel no longer provides `langgraph`, `dspy`, `examples`, or `regression`, and it changes the meaning of `all`, while `CHANGELOG.md` has no Unreleased entry. A clean `pip install -e ".[langgraph]"` exits 0 with `WARNING: assert-ai 0.2.0 does not provide the extra 'langgraph'`, then leaves `langgraph` uninstalled. Existing setup automation can therefore report a successful install and fail later at import time. Please either preserve safe compatibility aliases for a transition or make this an explicit breaking release change with the appropriate version/migration and changelog treatment. The vulnerable all-framework bundle does not need to be restored merely for compatibility.

Two smaller consistency issues:

- The new OpenAI auto-trace quickstart (`.[otel]` plus `examples/phoenix_auto_trace/requirements.txt`) emits `DependencyConflict: requested: "langchain_core >= 0.1.0" but found: "None"` because the `otel` extra still installs the LangChain instrumentor after LangChain moved to target-owned requirements. The framework instrumentor should move with the framework dependency or otherwise stop producing a conflict on the default OpenAI path.
- `scripts/render_trade_off.py:25-26` still says Matplotlib is a base `assert-ai` dependency, although this PR moves it to `scripts/requirements.txt`.

Other verification on this head was healthy: the full suite passed (`1454 passed, 22 skipped, 840 subtests`); viewer and website audits/builds passed with zero known npm vulnerabilities; `pip-audit` found no known vulnerabilities in the root and isolated requirement environments; the wheel and sdist passed `twine check`; the Travel Planner, Azure Doc QA, OpenAI auto-trace, scripts, and standalone Foundry-host requirements installed and imported; lockfiles use the public npm registry with SHA-512 integrity; and `uv lock --check` plus diff checks passed.

changliu2:
@jakepresent Ready for re-review at `59b9cd91`.

The commits after your `ed539e3` review address both blockers and both consistency issues:
- the Bank Manager documented install now resolves a compatible OpenAI/LiteLLM set and its clean install plus `pip check` is enforced in CI;
- the removed extras are documented as a planned 0.3.0 breaking migration and locked by the public-extra contract test;
- the LangChain instrumentor moved with the LangChain-owned example dependencies; and
- `render_trade_off.py` now points to the script requirements instead of calling Matplotlib a core dependency.

All current-head checks are green, and an independent exact-head review found no remaining blocker. It also verified that `mcp>=1.29.1` is published on public PyPI and receives the `<2` cap transitively from `langchain-mcp-adapters`; the earlier resolution concern came from a stale corporate mirror.

Please re-review the current head. We need #336’s dependency boundary merged before applying the approved article-faithful reset to #313.

jakepresent (changes requested):
The prior Bank Manager resolver conflict, 0.3 migration notes, instrumentor ownership, and script dependency text are fixed on exact head `8561aa2`. I found these remaining blockers:

1. **The non-GPT Bank Manager route uses the wrong OpenAI endpoint.** `examples/bank_manager_agent_control/agent.py:173-180` gives `ChatOpenAI` a base URL ending in `/models`, so it sends the request to `https://<resource>.openai.azure.com/models/chat/completions`. I confirmed that path with a mock HTTP transport. The current Microsoft migration contract for the OpenAI SDK requires `https://<resource>.openai.azure.com/openai/v1/chat/completions`: https://learn.microsoft.com/en-us/azure/foundry/how-to/model-inference-to-openai-migration. This breaks the advertised DeepSeek, Mistral, Llama, Phi, and Cohere branch. Please use the OpenAI v1 base path and add a transport-level request test; the current constructor test only asserts the incorrect URL.

2. **The documented LangGraph gallery evals do not install their target dependencies.** `examples/phoenix_auto_trace/README.md:101-113` presents two LangGraph configs as independently runnable after the quickstart, but `examples/phoenix_auto_trace/requirements.txt` contains only OpenAI and its instrumentor. In a clean Python 3.11 environment, the documented install passes `pip check`, then importing `examples.phoenix_auto_trace.travel_langgraph` fails with `ModuleNotFoundError: No module named 'langchain_core'`. The new CI job imports only `travel_openai`, so it misses this path. Please add or reference an adjacent LangGraph requirements set and exercise the configured target in CI.

3. **The Science Research Agent silently loses its required Tavily search capability.** `examples/science_research_agent/README.md:66-86` now installs base ASSERT only, while `tools.py:295-301` imports `TavilyClient`. Tavily is absent from the package extras and there is no adjacent requirements file. With caching disabled and `TAVILY_API_KEY` set, I reproduced `status: error` with `ModuleNotFoundError: No module named 'tavily'`. Because that exception becomes a tool result, the eval continues without its advertised public-web channel. Please add an adjacent requirements file and a clean install/tool check.

4. **Empty target responses still drop the new trace diagnostic from `inference_set.jsonl`.** `assert_ai/core/otel_session.py:278-285` attaches `span_validation` only to the final assistant interaction, while `assert_ai/stages/inference.py:366-388` records that interaction only when `content` is non-empty. My empty-response repro completed with `stop_reason="completed"` and one artifact event, but it persisted no `span_validation`. Please preserve raw-only diagnostic metadata on an always-written event or row and add an empty-response regression test.

5. **The supported `collector=` path now records a false missing-span failure.** `OTelTracedSession` documents `collector` as the preferred interface and stores it at `assert_ai/core/otel_session.py:66-105`, but `run_turn()` always reads `_exporter` at line 215. With `ListCollector([valid_span])`, direct validation returned `true`, while the session returned `valid:false`, zero trace events, and “No OpenTelemetry spans were captured.” Ignoring the collector is pre-existing, but this PR turns it into a newly persisted false failure. Please consume the supplied collector or avoid reporting its data as missing.

6. **The canonical banking CI instructions point to a nonexistent path in the linked repository.** `examples/bank_manager_agent_control/ci/README.md:11-16` points readers to `responsibleai/assert-ci-banking-demo`, then tells them to install `examples/bank_manager_agent_control/requirements.txt`. That path returns 404 in the external repository. Its actual root requirements still pin `assert-ai[acs,langgraph,otel,regression]==0.1.0`. Please coordinate the external-repo migration or make the in-repo text accurately describe which checkout each command applies to.

One smaller formatting issue also needs cleanup: the changed two-space indentation in `.claude/skills/run-assert-eval/SKILL.md:102-111`, `.github/prompts/run-assert-eval.prompt.md:39-42`, and `.cursor/rules/assert.mdc:67-70` breaks list/fence rendering. A CommonMark render puts the install command outside its fence and treats following preconditions as code until the next fence. Restoring the list indentation fixes it.

Verification on `8561aa2`: the full local suite passed with 1,454 tests, 24 skips, and 840 subtests; a synthetic merge with current `main` passed the focused 207-test check; the clean Bank Manager install and `pip check` passed; Python and npm vulnerability audits found no known vulnerabilities; and the viewer, website, package, and current GitHub checks passed.

jakepresent (approved):
Approved at exact head `75987cd`. I rechecked the six blockers from my prior review: the Foundry OpenAI v1 route, isolated LangGraph and Tavily dependency ownership plus CI coverage, empty-response trace diagnostics, injected collector consumption, external banking-repo setup wording, and Markdown rendering are all fixed.

The synthetic merge with current `main` passed 1,487 tests, 22 skips, and 840 subtests, and all current checks are passing. `PhoenixCollector` still needs repair, but its broken client construction and retrieval calls are inherited unchanged from `main`, it is not wired into the CLI/config path, and this PR improves the supported injected-collector path. I do not consider that inherited defect a merge blocker for this dependency-boundary change; it should be handled in a focused follow-up.

PR #340, opened 2026-09-01
Add workspace-scoped runtime path policy

AaronAspinwall123:
Introduce canonical workspace roots and typed containment for configs, inputs, outputs, and managed trees. Cover traversal, external roots, linked paths, and Windows path normalization with focused tests.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

## Changes

-
-

## Checklist

- [ ] Tests pass locally (`pytest` and/or viewer checks as applicable).
- [ ] Docs updated if behavior or public API changed.
- [ ] No secrets, credentials, or customer data committed.
- [ ] No breaking change, or a `CHANGELOG.md` entry is included.


changliu2 (changes requested):
Requesting changes on exact head `034fc7e` for three path-policy correctness issues:

1. Windows reserved device names are accepted as managed outputs. `resolve_output("NUL")` returns `<artifacts_root>\NUL`; writing to it succeeds against the DOS device but creates no artifact. Reject reserved/device path components before authorizing output paths.
2. Canonical Windows handling is applied only to containment checks, not the public APIs that follow them. `WorkspaceService.reference()` resolves relative inputs against the process CWD, an equivalent `\\?\C:\...` workspace path passes containment and then raises a raw `ValueError` at `relative_to`, and `require_managed_root()` rejects the equivalent extended path as an override. Anchor relative references to the workspace and use the same canonical representation for relative/equality operations.
3. Root validation permits invalid or aliased trees. An existing file is accepted as `config_root`, and an `evals` junction targeting `artifacts` collapses the config and managed-output roots so the same pathname is authorized for both. Reject existing non-directory roots and require config and managed-output roots to remain disjoint after canonical resolution.

changliu2 (approved):
Re-reviewed at `101fa787`. The Windows reserved-name and ADS cases, canonical root topology, workspace-relative and extended-path handling, case preservation, and snapshot-only link-check contract are addressed. All checks pass. Approving.

tangym (approved):

PR #342, opened 2026-09-02
fix(viewer): prevent results table and metric dropdown clipping

changliu2:
## Summary

- Keep the suite results table within its container with explicit column widths and wrapped metric labels.
- Truncate long run and target names instead of allowing them to expand the table.
- Position metric dropdown menus against the viewport so table overflow cannot clip their options.
- Retain horizontal scrolling as a fallback for narrower screens.

## Why

The clipping reported in #322 remains reproducible on current `main`. #324 contained an initial fix but closed without merging after the behavior was thought to be covered elsewhere. The suite results route still uses an `overflow-hidden` wrapper and non-wrapping metric controls, which can make the second metric and Total columns unreachable.

## Validation

- `npm ci`
- `npm run check` — 0 errors; existing accessibility warnings remain
- `npm run build`
- Browser reproduction with the 1200px results container: table `scrollWidth` equals `clientWidth`, and the second metric stays within the container
- Open second metric dropdown uses fixed positioning and remains fully within the viewport

## Scope

Viewer UI only. No evaluation artifacts, result values, configs, or runtime behavior changed.

### DIFF
--- a/viewer/src/lib/PrimerDropdown.svelte
+++ b/viewer/src/lib/PrimerDropdown.svelte
@@ -36,6 +36,40 @@
   let highlightedIndex = $state(-1);
   let triggerButton: HTMLButtonElement;
   let menuList: HTMLUListElement;
+  // Anchor the fixed-position menu to the trigger so it can escape table
+  // scroll and clipping contexts without inflating their scroll width.
+  let menuStyle = $state('');
+
+  const VIEWPORT_MARGIN = 8;
+  const TRIGGER_GAP = 4;
+  const MIN_MENU_WIDTH = 192;
+
+  function positionMenu() {
+    if (!triggerButton || !menuList) return;
+    const anchor = triggerButton.getBoundingClientRect();
+    const menu = menuList.getBoundingClientRect();
+
+    let left = anchor.left;
+    if (left + menu.width > window.innerWidth - VIEWPORT_MARGIN) {
+      left = anchor.right - menu.width;
+    }
+    left = Math.max(
+      VIEWPORT_MARGIN,
+      Math.min(left, window.innerWidth - menu.width - VIEWPORT_MARGIN)
+    );
+
+    let top = anchor.bottom + TRIGGER_GAP;
+    const overflowsBottom = top + menu.height > window.innerHeight - VIEWPORT_MARGIN;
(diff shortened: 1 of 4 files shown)


ango10:
Dropdown UI is now correct: image

ango10:
UI Review And Target Name Display: Chang and Alex examined PR #342 and the viewer’s handling of long run and target names, agreeing that Alex would review the proposed fix and consider making the related labels consistent. 04:54

- PR #342 Review: Chang reopened the previously closed UI change as PR #342 and asked Alex to review it. Alex recognized it as a previous PR and agreed to proceed with Chang’s version because the fix had appeared inconsistent across branches.
- Long Name Truncation: Chang explained that long run and target names are shortened instead of expanding the comparison table. Aaron recalled that this behavior replaced an earlier implementation where names overlapped, while Alex noted that user-defined callable names have no fixed format and can be arbitrarily long.
- Label Consistency: The discussion identified that the same underlying item can appear under different labels in the summary, comparison view, and cards. Alex and Chang considered using exactly consistent names, while retaining color coding as an additional visual aid; Alex will review the UI and suggest changes.
- Comparison View Context: Chang will send Alex richer run artifacts from the multi-compare example because Alex may not have an equivalent dataset. The artifacts will help Alex evaluate how run names, target names, and shortened labels are presented together.

Next action steps (Awaiting download permission from Chang 09/02/2026):

- Download Chang's "aiewf-blog-viewer-results-pr313 1.zip".
- Open "compare runs" view.
- Observe the 3 labels: Summary, By behavior category, and Test prompt are not aligned. e.g., 
acs-governed, chat_governed, chat_governed.
- Resolve this alignment issue.
image

ango10:
Commit https://github.com/responsibleai/ASSERT/pull/342/commits/4b1f811e54c56bb50a554bd1abe40863fd750950 resolves issue. See aligned labels + viewer here:
image

AaronAspinwall123 (approved):

changliu2:
Re-audited exact head `4b1f811`. The original clipping fix holds: the results table keeps every column reachable, the second metric remains inside the container, and its open dropdown stays fully within the viewport without creating a stray horizontal scrollbar.

The follow-up also consistently labels runs by run ID across the baseline selector, summary cards, behavior-category columns, and prompt/scenario sample panels, while retaining the target entrypoint as secondary text and in the tooltip. Viewer check/build and all required CI checks are green. Approving.

PR #344, opened 2026-09-07
add disability representation dataset

ahmedmagooda:
## Summary

- Added example dataset for disability representation harm

## Motivation / linked issue

NA

## Changes

- Added example dataset for disability representation harm

## Testing

NA

## Checklist

NA


changliu2 (changes requested):
Reviewed exact head `1d59297`. Two dataset correctness blockers remain:

1. The files use a schema that current ASSERT cannot consume. `policy.json` declares top-level `risk` / `behaviors`, while ASSERT requires `behavior` / `behavior_categories`; every `test_set.jsonl` row uses `kind`, `seed_id`, `concept`, and `factors`, while inference requires `type`, `test_case_id`, `behavior`, and `dimensions`. I validated these through the current schemas: the policy fails with three taxonomy errors, and inference rejects row 1 with `must declare type 'prompt' or 'scenario'`. There is no README, adapter, or other consumer explaining a separate archival format. Please convert the files to the supported artifact schema, or add and document the explicit loader/conversion path that makes this sample usable.

2. The assigned behavior coverage is incorrect for at least `test_set.jsonl:21-22`. Both rows are labeled `Severity 6: Explicit slurs/name-calling directed at disabled people`, but one asks for a comeback against a person harassing a disabled neighbor and the other asks for a line describing the drunk person harassing a wheelchair user. Neither request targets a disabled person with a slur, so half of this category's prompt coverage does not exercise the category definition. Please replace or relabel these cases and recheck category-to-prompt fidelity across the dataset.

The JSON is otherwise well formed, has 50 prompt and 50 scenario rows, covers all 26 declared categories, and contains no apparent credentials or personal contact data. The existing comments are empty/non-blocking, but the two issues above block approval.

